PR #30, opened 2022-09-21
chore: add note to example readmes about building main package

yusefnapora:
This just adds a couple lines to the example READMEs to make sure people build the main `w3ui` packages before running the examples.  Don't know if there's an issue for it, but closes @jchris's slack ping :smile:


jchris (approved):
this is good!

PR #103, opened 2022-10-26
WIP: Implement upload/* caps and make upload file invoke upload/add along with store/add.

ice-breaker-tg:
WIP
This is to address breaking change in list API / upload capabilities being added to w3up.

It is nearly complete, but may require some tweaks.

PR #104, opened 2022-10-27
fix: uploads listing

alanshaw:
This PR fixes the uploads listing by calling the `upload/list` capability.

After a chunked upload is complete, the uploader now calls `upload/add` to register an "upload" and associate multiple CAR files together.

`upload/list` capability will list out the uploads registered using the `upload/add` call.

There's no breaking API changes but existing uploads will no longer be displayed in the list since they were not registered using `upload/add`.

supersedes https://github.com/web3-storage/w3ui/pull/103


alanshaw:
I'm going to merge this because we're short on time.

PR #121, opened 2022-11-09
fix: Use cloneNode for vanilla templates

thewtex:
When extending the file-upload to reset the element after initial upload, this addresses the [DocumentFragment
pitfall](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template#avoiding_documentfragment_pitfall).


thewtex:
Deployed here: https://cmake-w3-externaldata-upload.on.fleek.co/

Thank you for your work on this project! :pray:

PR #119, opened 2022-11-03
feat!: integrate new ucanto and upload-client

alanshaw:
This PR makes breaking changes to core and framework components to integrate the access agent (+ucanto 0.9) and the new upload client.

resolves #127

BREAKING CHANGE: core and framework components have changed considerably.

TODO: be more explicit about changes

PR #125, opened 2022-11-16
fix: update return types of providers

Codigo-Fuentes:


alanshaw:
This will land in https://github.com/web3-storage/w3ui/pull/119 - thanks for the fix.

PR #129, opened 2022-11-28
fix: update vanilla sign-up example

yusefnapora:
This just updates the vanilla sign up/in example to use the new stuff from #119.

I've got the file upload example nearly done as well, but need to wrap things up for the day, so I'll finish it tomorrow.

@alanshaw this PR is targeting the branch from #119, so feel free to pull it in there if you want. Otherwise I'll retarget this one after that one merges.


travis (approved):
looks good to me - not entirely sure we should check in the staging service details, and if so probably nice to stash them somewhere we can use them in all examples, but not sure exactly where

PR #130, opened 2022-11-29
fix: update vue examples

travis:
I got all these working in a basic way, but had trouble passing the staging service principals and connections through to the providers to test against staging - I think this is the right way to pass them through in Vue:

Screen Shot 2022-11-28 at 5 22 51 PM

but Vue complains like this:

Screen Shot 2022-11-28 at 5 24 08 PM

and the servicePrincipal and connection are undefined inside KeyringProvider - @alanshaw am I just bad at Vue or do we need to do something special to make sure servicePrincipal and connection are defined as "props" in the Providers?


travis:
I got all these working in a basic way, but had trouble passing the staging service principals and connections through to the providers - I think this is the right way to pass them through in Vue:

Screen Shot 2022-11-28 at 5 22 51 PM

but Vue complains like this:

Screen Shot 2022-11-28 at 5 24 08 PM

and the `servicePrincipal` and `connection` are undefined inside `KeyringProvider` - @alanshaw am I just bad at Vue  or do we need to do something special to make sure `servicePrincipal` and `connection` are defined as "props" in the Providers?

travis:
added a commit that fixes the issue above after a chat with @alanshaw - I still haven't been able to confirm this is working end-to-end as I'm getting a 500 from the execute API:

Screen Shot 2022-11-29 at 12 21 07 PM

This PR should be fine to merge anyway - I believe @yusefnapora was also seeing a 500 so we can track that down separately.

alanshaw (approved):

PR #131, opened 2022-11-30
fix: update solid examples

travis:
This should do the trick to update the various examples, but I haven't tested end to end because I kept getting 500s from the uploads service.


alanshaw (approved):
🎉

PR #135, opened 2022-11-30
chore: update docs for new ucanto integration

yusefnapora:
This updates the markdown docs based on the changes in #119.

I _think_ I got everything, but @alanshaw should probably take a look :)


docs/solid-keyring.md
```diff
@@ -14,87 +14,44 @@ import * as SolidKeyring from '@w3ui/solid-keyring'
...
+
+### `useKeyring`
 
 ```ts
-const [state, actions] = useAuth()
+const [state, actions] = useKeyring()
 ```
 
-Hook to allow use of the [`AuthProvider`](#authprovider) value. The value returned is an `AuthContextValue`:
+Hook to allow use of the [`AuthProvider`](#authprovider) value. The value returned is a `KeyringContextValue`:
```
alanshaw, line 48:
```suggestion
Hook to allow use of the [`KeyringProvider`](#keyringprovider) value. The value returned is a `KeyringContextValue`:
```

alanshaw (approved):
🥳 Awesome thank you!

PR #136, opened 2022-12-01
feat: "Headless" UI components

travis:
"Headless" UI React components

`react-{keyring|uploader|uploads-list}` now all provide "headless" UI components modeled after https://headlessui.com/

The Big Idea with these is that they provide functionality and look like regular UI components but don't force developers into any particular markup or styling choices. They're very useful "lego" building blocks that work together to encapsulate the details of interacting with our services and let developer focus on building UI however they prefer to do so. Properties like `className` are passed along to the underlying component. 

As of this writing, it's not possible to change the markup tag for some of the headless components - this should be fixed (possibly in this PR?) by adding an `as` property to these components similar to HeadlessUI (https://headlessui.com/react/menu#menu)


packages/react-ui/src/Uploader.jsx
```diff
@@ -0,0 +1,116 @@
...
+      setDataCid(cid)
+    } catch (err) {
+      console.error(err)
+      setError(err)
+    } finally {
+      setStatus('done')
+    }
+  }
+  return (
+    <UploaderContext.Provider value={{ storedDAGShards, file, setFile, dataCid, status, error }}>
```
jchris, line 33: 👍

packages/react-ui/src/Uploader.jsx
```diff
@@ -0,0 +1,116 @@
...
+      {(status === 'uploading') ? (
+        <Uploading file={file} storedDAGShards={storedDAGShards} />
+      ) : (
+        (status === 'done') ? (
+          error ? <Errored error={error} /> : <Done file={file} dataCid={dataCid} storedDAGShards={storedDAGShards} />
+        ) : (
+          <form onSubmit={handleUploadSubmit} {...props}>
```
jchris, line 40: should this use Uploader.Form?

packages/react-ui/src/Uploader.jsx
```diff
@@ -0,0 +1,116 @@
...
+    <h1>⚠️ Error: failed to upload file: {error.message}</h1>
+    <p>Check the browser console for details.</p>
+  </div>
+)
+
+const DefaultDone = ({ file, dataCid, storedDAGShards }) => (
+  <div>
+    <h1>Done!</h1>
+    <p>{dataCid.toString()}</p>
+    <p><a href={`https://w3s.link/ipfs/${dataCid}`}>View {file.name} on IPFS Gateway.</a></p>
```
jchris, line 76: should we start defaulting to `https://${dataCid}.ipfs.w3s.link/` form instead of `https://w3s.link/ipfs/${dataCid}` to save a redirect? @vasco-santos @alanshaw

jchris (commented):
Thanks for starting this!

packages/react-ui/src/Uploader.jsx
jchris: but https://filecoinproject.slack.com/archives/C02BZPRS9HP/p1669839337281549?thread_ts=1668123991.879979&cid=C02BZPRS9HP

packages/react-ui/src/Uploader.jsx
travis: oh, actually this should be inserting `children` here rather than providing any UI - one of the goals of this particular component is to avoid concrete HTML entirely so that consumers have maximum markup freedom - as currently written the `SimpleUploader` example doesn't work at all, lemme update..

packages/react-ui/src/Uploader.jsx
vasco-santos: @jchris cannot open slack thread. But yes, I would really like that we start to recommend people to use subdomain directly.

alanshaw (changes requested):
I love the "headless" components. They need documentation though please.

Can we split the `react-ui` stuff into a separate PR? I think it's a good idea to provide these components...but I'm just not 100% on the name "react-ui"...it sounds too general. I think it'll just become a pile of unrelated components rather than a focused set. Perhaps they could just be exported from the existing packages under `/ui` e.g. `@w3ui/react-uploader/ui` or just as another export `import { UI } from '@w3ui/react-uploader'` && `<UI.SimpleUploader />`?

travis:
> They need documentation though please.

yep agreed! I can take a first pass and then maybe @yusefnapora can come through and fix/improve what I add?

> Perhaps they could just be exported from the existing packages under /ui 

Yep I considered this but 

a) then there's no clear place for the "dropin" component to live
b) people who just want to use the Headless components get all the deps required for the Customizable components (tree shaking should make this not a big deal for many users, but still not ideal
c) I have a soft preference for sticking to one level of abstraction per package (ie, it feels like a slightly better developer experience to know that all of the components in the "headless" packages follow the "headless" conventions)

One solution that would solve all of these is to introduce`@w3ui/{uploader,uploads-list,keyring}-ui` packages or something similar, but that feels like an explosion of packages once we apply the pattern to Vue and Solid as well...

I'm definitely down to find a different name, what works best probably depends on how we want to organize these components.

In any case, I'll move the `react-ui` related stuff to its own PR and get the headless components in ASAP

travis:
ok this PR is dramatically simplified, only includes the Headless components. react-ui has been moved to #208 where we can figure out the naming

alanshaw (changes requested):
This looks awesome

packages/react-keyring/src/Authenticator.tsx
```diff
@@ -0,0 +1,132 @@
...
+ * Designed to be used by Authenticator.Form, Authenticator.EmailInput
+ * and others to make it easy to implement authentication UI.
+ */
+export function Authenticator (props: any): JSX.Element {
+  const [state, actions] = useKeyring()
+  const { createSpace, registerSpace } = actions
+  const [email, setEmail] = useState('')
+  const [submitted, setSubmitted] = useState(false)
+
+  const handleRegisterSubmit = useCallback(async (e: React.FormEvent<HTMLFormElement>) => {
```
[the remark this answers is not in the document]
travis: >  In both cases they add scope for subtle bugs since you have to manually specify dependencies.

Yep totally agree with this AND in this case because we're developing components designed to be used by outside developers I think it's more important to be careful with this stuff - this is a good example - if we don't use `useCallback` the `handleRegisterSubmit` function will be recreated on every render, triggering unnecessary re-renders in downstream components - in this case, because we pass this callback to the Context, it will potentially trigger unnecessary re-renders in any component that calls `useAuthenticator` (possibly even if they don't use `handleRegisterSubmit`, though I'm not 100% sure about that)

tldr: I agree with you philosophically but think we should use `useCallback` consistently for all callbacks in these libraries.

packages/react-keyring/src/Authenticator.tsx
```diff
@@ -0,0 +1,132 @@
...
+  )
+}
+
+/**
+ * Form component for the headless Authenticator.
+ *
+ * A `form` designed to work with `Authenticator`. Any passed props will
+ * be passed along to the `form` component.
+ */
+Authenticator.Form = function Form (props: any) {
```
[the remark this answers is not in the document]
travis: yes! I've actually overhauled these types a bit in the `as` property work I'm doing in #236 so I'd like to punt this particular task there, but I'd also like to get that work merged before we do a release of these components, so worth taking a look over there and seeing if the typing is a bit more to your liking

packages/react-uploads-list/src/UploadsList.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * Button that loads the next page of results.
+ *
+ * A 'button' designed to work with `UploadsList`. Any passed props will
+ * be passed along to the `button` component.
+ */
+UploadsList.NextButton = (props: any) => {
+  const [, { next }] = useContext(UploadsListComponentContext)
+  const onClick = useCallback(function onClick (e: React.MouseEvent) {
```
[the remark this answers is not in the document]
travis: good question, I prefer arrow too, will fix

packages/react-uploader/src/Uploader.tsx
```diff
@@ -0,0 +1,138 @@
+import React, { useContext, useMemo, createContext, useState } from 'react'
+import { Link, Version } from 'multiformats'
+import { CARMetadata, UploaderContextState, UploaderContextActions } from '@w3ui/uploader-core'
+import { useUploader } from './providers/Uploader'
+
+export type UploaderComponentContextState = UploaderContextState & {
+  /**
+   * A string indicating the status of this component - can be 'uploading', 'done' or ''.
```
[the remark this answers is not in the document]
travis: agreed on all counts! switched the API to:

```
export enum Status {
  Idle = 'idle',
  Uploading = 'uploading',
  Failed = 'failed',
  Succeeded = 'succeeded'
}
```

packages/react-uploader/src/Uploader.tsx
```diff
@@ -0,0 +1,138 @@
...
+/**
+ * Input component for the headless Uploader.
+ *
+ * A file `input` designed to work with `Uploader`. Any passed props will
+ * be passed along to the `input` component.
+ */
+Uploader.Input = (props: any): JSX.Element => {
+  const [, { setFile }] = useContext(UploaderComponentContext)
+  return (
+    <input type='file' onChange={e => setFile(e?.target?.files?.[0])} {...props} />
```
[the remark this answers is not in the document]
travis: yep! great call

packages/react-keyring/src/Authenticator.tsx
```diff
@@ -0,0 +1,132 @@
...
+  const [submitted, setSubmitted] = useState(false)
+
+  const handleRegisterSubmit = useCallback(async (e: React.FormEvent<HTMLFormElement>) => {
+    e.preventDefault()
+    setSubmitted(true)
+    try {
+      await createSpace()
+      await registerSpace(email)
+    } catch (err: any) {
+      throw new Error('failed to register')
```
alanshaw, line 73: Do not smother!

```suggestion
      throw new Error('failed to register', { cause: err })
```

Separately, we should probably not throw this but instead expose an error state variable. This function will likely be passed directly to the `onSubmit` handler for the form so there's no opportunity to catch it. Can be a follow up ;)

packages/react-keyring/src/Authenticator.tsx
```diff
@@ -0,0 +1,132 @@
...
+/**
+ * Input component for the headless Uploader.
+ *
+ * An email `input` designed to work with `Authenticator.Form`. Any passed props will
+ * be passed along to the `input` component.
+ */
+Authenticator.EmailInput = function EmailInput (props: any) {
+  const [{ email }, { setEmail }] = useAuthenticator()
+  return (
+    <input type='email' value={email} onChange={e => setEmail(e.target.value)} {...props} />
```
alanshaw, line 110: Any time we receive arbitrary props we want to spread them first to make sure the user doesn't override anything we don't want them to override. There a few more instances of this.

alanshaw (approved):

packages/react-keyring/src/Authenticator.tsx
travis: ah yea - my thinking here was that it would be better to give the user the ability to override this stuff if needed, but it's definitely a tradeoff between flexibility and usability - if we opt to let them override we open up more space for accidentally overriding things that are important to the functionality of our component, and if we opt not to we might force the user to abandon/reimplement these components themselves in some cases.

concrete example is the `type` property - we don't technically need that to be `email` for this component to work in a reasonable way, but it is a useful default.

I'm not really sure of the right tradeoff across the board, so I'm inclined to go with your solution and be consistent about always spreading first. If we find users are asking for the ability to override properties, we can change this in the future without breaking backwards compatibility.

packages/react-keyring/src/Authenticator.tsx
travis: ah great call - I think TypeScript complained when I tried to pass `cause` previously, but I'll take another crack at it.

totally agree re: not throwing this, I'll get that into a followup!

packages/react-keyring/src/Authenticator.tsx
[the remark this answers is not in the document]
olizilla: I don't have a strong opinion about this. I'm happy to go with @travis' instinct here. If react users are going to complain at us for not doing it cos they all do, and it's not too onerous to apply it consistently then let's go! 

Alternatively, we could drop these optimisations, and wait for users (or better yet, our perf tests) to flag where we really need this. Optimising existing components can be a great "first contribution" task where folks want to get involved.

packages/react-keyring/src/Authenticator.tsx
alanshaw: > I'm inclined to go with your solution and be consistent about always spreading first

That not exactly what I intended to propose above. It's only the things we don't want users to override that should _not_ be overridable. As you correctly identified we might set properties that are safe to be overridden. The concern was more (in this instance) that if we allow `value` or `onChange` to be overridden this component cannot operate correctly.

travis: oh got it got it - ok, I agree with that completely! I do still think it's good to start from a place of not letting them override anything until we document what is overridable so that we don't have people depending on behavior we decide to deprecate (ie, going from overridable to not overridable is a breaking change because people might be depending on being able to override something, going from not overridable to overridable isn't) so I'm happy this landed where it did, but I'll feel free to move some props to after the spread in the future!

PR #137, opened 2022-12-02
chore: move Uploader into react-uploader package

travis:
I think it makes more sense to include these "headless UI" style components in the @w3ui/react-* packages rather than bundling them together in the react-ui package. It feels more intuitive given the naming - if we do want to have the "no UI" style components in a separate package I'd advocate for callling that package `react-uploader-provider` or `react-uploader-api` or something, but given how much code this is and the fact that tree shaking is pretty good, it seems ok to me to bundle them together.

the react-ui package will be reserved for higher level "customizable UI" style components and can now have opinions about the right way to do customization at the top level rather than having two different flavors of customizability in a single package.

@alanshaw @cmunns curious what you think about this - happy to move the uploader back if we decide that's a better structure

### DIFF
--- a/examples/react/ui/src/ContentPage.js
+++ b/examples/react/ui/src/ContentPage.js
@@ -1,6 +1,6 @@
 import React from 'react'
-import { Uploader, SimpleUploader } from '@w3ui/react-ui'
-import { UploaderProvider, useUploader } from '@w3ui/react-uploader'
+import { SimpleUploader } from '@w3ui/react-ui'
+import { Uploader, UploaderProvider, useUploader } from '@w3ui/react-uploader'
 import { Prism as SyntaxHighlighter } from 'react-syntax-highlighter';
 import { a11yDark } from 'react-syntax-highlighter/dist/esm/styles/prism';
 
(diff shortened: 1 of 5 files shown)


cmunns:
My only point of friction on this is somebody who previously installed the package as-is later wanting to improve upon it with a few customizations will need to rip out previous work and re-install as opposed to ad-hoc changes. Other than that I don't have a strong opinion on separating the packages this way.

cmunns:
Actually after a little more thought I do agree that it makes sense to split out the package that is "styled" into a *-ui. It does help imply what you are importing and lets the functional components stand by themselves in a cleaner way.

travis:
> My only point of friction on this is somebody who previously installed the package as-is later wanting to improve upon it with a few customizations will need to rip out previous work and re-install as opposed to ad-hoc changes. 

This is a good point though! Making a mental note to make sure to document this explicitly to help folks understand how to make this transition smoothly - I do think that this is a cleaner way to separate things, but "moving down the abstraction stack" here is definitely a little more complicated  - developers will need to a) find where the components that underlie the current component they're working on live and b) import and use them - I think in both cases looking at the source code will be super helpful, so we should make sure to have links to the GitHub source in documentation of the "customizable UI" components.

travis:
I'm going to merge this since it's just a merge into my w3ui-ui feature branch, but still happy to revisit this decision

PR #140, opened 2022-12-05
First stab at css custom properties and class convention for multi-uploader

cmunns:
also swapped the spinner for something on-brand.
Screenshot 2022-12-05 at 1 57 48 PM


examples/react/multi-file-upload/src/style.css
```diff
@@ -0,0 +1,206 @@
...
+  --w3ui-uploader-primary: salmon;
+  --w3ui-uploader-primary-hover: rgb(250, 136, 124);
+}
+
+.w3ui-app {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+  min-height: 100vh;
```
jchris, line 26: I end up having to remove this on apps that scroll. maybe there is a way to be more flexible by default?

cmunns: Yes. ultimately we don't want this in here, but it makes it less ugly as we test these things out.

travis:
this is super cool! the APIs here make sense to me - I could see adding variables for a few more things but am inclined to let that shake out of usage and requests from users so this looks great for now - it'd definitely be great to see a couple different examples of designs that are possible with this set of variables ([w3ui zen garden](http://www.csszengarden.com/) anyone? 

I'd be happy to see this merge if @alanshaw is happy!

alanshaw:
It's fine by me if you want to pull tachyons out of the examples. I think it's a good idea not to use a framework for them 😁

Please can we retain consistency though, and use single quotes everywhere - standard js should be enforcing this already, not sure how this is not being flagged in the checks.

Also in the name of consistency we need to propagate these styles to the examples in the other frameworks. Happy to merge this as is but we need a follow up asap - I'm hoping to merge and release `next` by the end of the week.

travis:
ok great - agreed!

> Please can we retain consistency though, and use single quotes everywhere - standard js should be enforcing this already, not sure how this is not being flagged in the checks.

oo I'm probably guilty of double quote usage - noted and will update as I see places where I've added double quotes

> Also in the name of consistency we need to propagate these styles to the examples in the other frameworks.

Agreed! fwiw I'm happy to hold off on merging this to `next` until after we do that release, but @cmunns if you have time to update the other examples to this style before that merge it'd be cool to get this in!

travis:
I ported this over to the "customizable" UI components in https://github.com/web3-storage/w3ui/pull/240 - closing for now @cmunns next time you're working on this stuff I'd love to review to make sure I didn't lose anything

PR #141, opened 2022-12-06
feat: UploadsList components

travis:
first pass at UploadsLists headless components. needs an example in react-ui to prove out

two main questions for reviewers:

1) any thoughts on the API for this component? feel free to pass on evaluating this until I have an example/docs
2) is it worth adding a "prev" function to the UploaderProvider so that we could build out a real basic "paging" component? I don't think we'll be able to add current page/total number of pages controls and indicators without more support from the backend service, but just forward and backward is better than a one way street
3) otoh, should I maybe add an infinite scroll list component to show off what we can do with the current API?

TODO:
- [ ] get feedback on API
- [x] add example in react-ui
- [x] add `prev` upstream to allow us to build basic pager component (added #143 for this)
- [x] add infinite scroll list subcomponent (added #144 for this)

### DIFF
--- a/packages/react-ui/src/SimpleUploader.tsx
+++ b/packages/react-ui/src/SimpleUploader.tsx
@@ -1,6 +1,6 @@
-import React, { useContext } from 'react'
+import React from 'react'
 import { CARMetadata } from '@w3ui/uploader-core'
-import Uploader, { UploaderContext } from '@w3ui/react-uploader';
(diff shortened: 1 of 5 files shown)


travis:
ok I'm going to merge this since it's a sub-PR and will get reviewed with #136 - I'm confident enough after chatting with folks about this that it's the right thing that it doesn't need a separate review

PR #151, opened 2022-12-07
fix docs link from uploads-list to keyring page

natevw:
at least in certain contexts (specifically I was viewing from the URL <https://github.com/web3-storage/w3ui/blob/next/docs/react-uploads-list.md> via the `next` branch) the useKeyring documentation link led to a 404 when clicked. Adding the `.md` extension seems to be all that's needed.


travis (approved):
lgtm! @yusefnapora is this just a typo or is there supposed to be some .md extension magic going on here?

PR #152, opened 2022-12-07
feat: Add Authenticator and SimpleAuthenticator

travis:
add "headless" Authenticator component and use it to build a "customizable" SimpleAuthenticator that works the way the Authenticator components in our examples work

separate PR against `feat/w3ui-ui` to make feedback easier - this component will replace the `Authenticator` components we've copied into a number of different examples

this is the last major React "customizable" UI component - if we're happy with this we should probably push forward with Vue and SolidJS versions of each of them - hopefully more straightforward now that the architecture/design work has been done


travis:
I'm gonna merge this one too since it will be reviewed as part of #136 🤠

PR #155, opened 2022-12-09
feat: W3Upload "drop in" component

travis:
combine all three simple components to create an easy to use dropin component and add W3Upload to the react ui example app


travis:
merging because review will still be done in the target branch

PR #191, opened 2022-12-15
fix: run babel on .js files as well as .ts

### LINKED ISSUE #160: codesandbox examples are not compiling
I believe because we now have `node_modules` that include private class properties.

yusefnapora:
This adds `.js` to the set of file extensions processed by babel.

I'm hoping that this will fix #160, since after running `npm run build` with this change I no longer see any class private members in `packages/react-keyring/build/esm/index.js`, which is where the example was failing before.

The downside of this change is that the rolled-up `index.js` grew from 198kb to 300kb for react-keyring. I assume the other modules are affected as well.

I also haven't been able to test this in codesandbox, since I don't know how to make it depend on an unreleased npm package (or if that's even possible).

### DIFF
--- a/rollup.config.ts
+++ b/rollup.config.ts
@@ -30,7 +30,7 @@ const umdDevPlugin = (type: 'development' | 'production'): Plugin =>
 const babelPlugin = babel({
   babelHelpers: 'bundled',
   exclude: /node_modules/,
-  extensions: ['.ts', '.tsx']
+  extensions: ['.ts', '.tsx', '.js']
 })
 
 export default function rollup (options: RollupOptions): RollupOptions[] {


yusefnapora:
Hey, what do you know? Turns out the codesandbox github app automatically pulled in the `react-keyring` from this branch, and it seems to be working in the bot-linked examples above.

travis:
this seems... fine?! clients of this library should be using tree-shaking, so hopefully package size increases won't matter, so that seems ok too? 

stepping back, I'm interested in overhauling the rollup configs at some point to allow us to rollup-build packages separately so hopefully we can optimize the size of various things in this package when we do that, which puts me firmly in the "shipit" camp - would love @alanshaw to take a look as well before merge!

yusefnapora:
Since Alan's off on holiday, I'll go ahead and merge this before I forget about it :)

PR #192, opened 2022-12-16
Fix/uploads list data

travis:
the data passed back from `list` has changed, so these
examples need to be updated. as I was digging in I realized
the examples don't seem to be working and I think we need
to dig deeper into the packages to understand why, so
I'm checkpointing here before I dig into that


travis:
superceded by #195 and #198

PR #195, opened 2022-12-16
fix: update react uploads list example

travis:
one slightly weird thing about this is that I realized we need to manually call `next` to get the next page of results - not sure if that's intentional and we might want to add that to the `react-uploads-list` package for developer convenience


travis:
@natevw this is the PR with the react example fix

yusefnapora (approved):
LGTM :)

jchris (approved):
lgtm

PR #198, opened 2022-12-19
fix: update vue examples

travis:
* make sure to loadAgent now that the package doesn't
* update uploads list for new data uploads list data format

### DIFF
--- a/examples/vue/file-upload/src/components/Authenticator.vue
+++ b/examples/vue/file-upload/src/components/Authenticator.vue
@@ -1,5 +1,5 @@
 <script>
-import { defineComponent } from 'vue'
+import { defineComponent, inject } from 'vue'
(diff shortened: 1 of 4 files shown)


travis:
@alanshaw this PR assumes that this commented-out line in the vue keyring was commented intentionally - is that the case? happy to uncomment and adjust this PR if that makes more sense https://github.com/web3-storage/w3ui/blob/main/packages/vue-keyring/src/providers/Keyring.ts#L115

yusefnapora (approved):
LGTM, assuming that removing `loadAgent` from the vue keyring package was intentional :)

I think it's fine to merge this now to get things working and update later if we want to do the loadAgent in the keyring package after all.

travis:
> LGTM, assuming that removing `loadAgent` from the vue keyring package was intentional :)

great! I've filed #201 to track the `loadAgent` question/improvement

PR #203, opened 2022-12-22
fix: get react file-upload example working with vite

travis:
`create-react-app` projects don't have the proper loaders to handle loading typescript in dependencies. `vite` is used in our other examples so it makes sense to me that we'd use it for react as well.

putting this in a separate PR before I update all the other examples to get feedback before committing to this approach, and to avoid too many merge conflicts with the efforts underway to update examples by @cmunns and @natevw


jchris:
LGTM, only question is does it make it more confusing for beginners to get hacking on it? Probably it's fine

jchris (approved):
👍

natevw:
Overall looks good to me too. +1 on giving the JSX files an actual `.jsx` extension, and cleaning up the public folder.

My one note is I'm not sure why the .gitignore file needed so many changes just for this commit and/or at this level of the repo **but** that isn't a blocker and not worth bikeshedding imo (says the guy who brought it up, sorry… 😂 ).

travis:
ok amazing - @alanshaw I'd like your signoff before I convert the rest of the examples to vite - happy to add that to this PR if we're all aligned

travis:
> only question is does it make it more confusing for beginners to get hacking on it?

I don't think so honestly - tbh I would never recommend `create-react-app` to beginners these days, and we already use vite for all the other examples. 

more detailed CRA vs Vite breakdown here: https://semaphoreci.com/blog/vite

alanshaw (approved):
👍 happy to use newer and better tooling and it's nice to have consistency with other examples anyway.

PR #208, opened 2023-01-04
feat: Customizable UI components

travis:
I've introduced a new package called react-ui that uses the headless components to build higher level components with some markup and styling. The markup is not customizable, but @cmunns has been proving out how we can make the styling customizable using CSS variables in https://github.com/web3-storage/w3ui/pull/140

react-ui has 3 components that simply wrap the underlying react-{keyring|uploader|uploads-list} packages and one that brings them all together into a component that developers should be able to simply drop into their applications and start uploading files. I'm conflicted on whether we should merge this now and iterate on main or separate the headless UI components into their own PR and merge react-ui later once we've figured out the CSS variable story - looking for feedback!


travis:
continuing the conversation from https://github.com/web3-storage/w3ui/pull/136#pullrequestreview-1236015303 and https://github.com/web3-storage/w3ui/pull/136#issuecomment-1371489402 - @alanshaw I'm still leaning toward a monolithic `react-ui` component for now for all the "customizable" components but open to other ideas per my comment in #136

examples/react/ui/src/ContentPage.jsx
```diff
@@ -0,0 +1,162 @@
...
+        w3ui provides React components at four different levels of
+        abstraction, each of which builds upon the last. In order they are:
+      </p>
+      <h3>1. No UI</h3>
+      <p>
+        Maximum flexibility, React Contexts and hooks only:
+      </p>
+      <SyntaxHighlighter language='jsx' style={a11yDark}>
+        {`
+function NoUIUploadComponent() {
```
olizilla, line 74: Ooh! Could toString on the existing functions here to avoid having it twice!

```suggestion
${NoUIUploadComponent.toString()}
```

olizilla:
Please could you add screenshots to PRs that add new chunks of UI, so that we can give feedback and compare what we see when we run it locally with what you expect to catch and surprises.

examples/react/ui/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+  ```sh
+  npm install
+  npm run build
+  ```
+
+- Change to this example directory and run the example:
+
+  ```sh
+  cd examples/react/template
```
olizilla, line 22:
```suggestion
  cd examples/react/ui
```

examples/react/ui/README.md
```diff
@@ -0,0 +1,24 @@
...
+  ```sh
+  git clone https://github.com/web3-storage/w3ui
+  cd w3ui
+  ```
+
+- Install dependencies and build:
+
+  ```sh
+  npm install
+  npm run build
```
olizilla, line 16:
```suggestion
  pnpm install
```

examples/react/ui/index.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <link rel="icon" href="/favicon.ico" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>W3UI File Upload Example App</title>
```
olizilla, line 7:
```suggestion
    <title>W3UI React Components</title>
```

examples/react/ui/README.md
```diff
@@ -0,0 +1,24 @@
...
+  ```
+
+- Change to this example directory and run the example:
+
+  ```sh
+  cd examples/react/template
+  npm start
```
olizilla, line 23:
```suggestion
  pnpm start
```

examples/react/ui/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "@w3ui/example-react-ui",
+  "private": true,
+  "version": "0.0.0",
+  "type": "module",
+  "scripts": {
+    "dev": "vite",
```
olizilla, line 7:
```suggestion
  "scripts": {
    "start": "vite",
    "dev": "vite",
```

travis:
This package also introduces an example app that demonstrates how to use the different abstractions in this repository:

Screen Shot 2023-01-10 at 2 37 10 PM
Screen Shot 2023-01-10 at 2 37 17 PM
Screen Shot 2023-01-10 at 2 37 25 PM
Screen Shot 2023-01-10 at 2 37 31 PM

travis:
#255 now has some important updates and bugfixes to the components in this PR - I'm going to backport them here since this PR is not ok to merge without them

olizilla:
Minor but I think we should introduce our layers the other way around in the example app... so Drop- In UI is first. As an impatient developer, the drop-in UI is the simplest to understand and use. If it does what I need I'm done! I can stop reading... but maybe i'm curious, _ooh can i make it cornflower blue?_ oh rad there is a Customizable UI too... _but english isn't my first language, can i change the lables_ oh rad i can rebuild the UI as I like from the Headless UI!

I'm not convinced yet that we need to introduce 4 levels here. I see it more as 2 levels like "Drop in components that you can customize the lookNfeel of, or Headless components that you can use as utils to build what you need". I don't want to have to think too hard to start using w3ui!

olizilla:
We need to resolve the conversation from https://github.com/web3-storage/w3ui/pull/136#pullrequestreview-1236015303 included here to make it easier to follow:

> Can we split the react-ui stuff into a separate PR? I think it's a good idea to provide these components...but I'm just not 100% on the name "react-ui"...it sounds too general. I think it'll just become a pile of unrelated components rather than a focused set. Perhaps they could just be exported from the existing packages under /ui e.g. @w3ui/react-uploader/ui or just as another export import { UI } from '@w3ui/react-uploader' && <UI.SimpleUploader />?
– @alanshaw ([source](https://github.com/web3-storage/w3ui/pull/136#pullrequestreview-1236015303))

> Yep I considered this but
>
> a) then there's no clear place for the "dropin" component to live
b) people who just want to use the Headless components get all the deps required for the Customizable components (tree shaking should make this not a big deal for many users, but still not really ideal)
c) I have a soft preference for sticking to one level of abstraction per package (ie, it feels like a slightly better developer experience to know that all of the components in the Headless packages follow the Headless conventions)
>
> One solution that would solve all of these is to introduce@w3ui/{uploader,uploads-list,keyring}-ui packages or something similar, but that feels like an explosion of packages once we apply the pattern to Vue and Solid as well...
– @travis ([source](https://github.com/web3-storage/w3ui/pull/136#issuecomment-1371489402))

_no doubt I will retred old ground here; bare with me i need to arrange my thoughts!_

tanstack was the inspiration for the existing packages. The terminology used there is headless vs component based libraries. https://tanstack.com/table/v8/docs/guide/introduction#component-based-libraries-vs-headless-libraries

tanstack implements UI agnostic "core" libs and adapters that map the core state handling logic to each ui libraries preferred way... e.g @tanstack/react-table provides react hooks.

tanstack deliberately avoids offering components that provide markup, so we are adding another dimension to our offering with our customisable components and headless ones... it's good, but we have to take care when explaining it. 

I understand that https://github.com/tailwindlabs/headlessui is an inspiration for the `react-ui` proposed here, which produces some DOM, but aims to make them customisable with tailwind patterns.

I note that we're likely to end up with top level packages for the headless implementations, and then the same again as a tree of components nested under `packages/react-ui`, but as @travis notes it does let us add compound components like the drop-flavour to that package without forcing us to add a new package for everything. We should review this once we've done a couple of iterations on w3console and see if we are happy with it.

To nudge this forwards I suggest we proceed with @travis's layout with minor tweaks to the naming:

- `packages/react-ui` -> `packages/react` in the source tree
- `@w3ui/react-ui` becomes `@w3ui/react` in the package name
- drop the `Simple` prefix from the customisable component names

and we can (maybe, later) re-export all the headless react components from `@w3ui/react/index` so that developers who want the easiest DX can import anything from `@w3ui/react`... while others who don't trust the tree-shake and just want the bare minimum can install the relevant headless package directly.

I lean towards landing this sooner rather than later with those minimal changes and see how it goes. We'll iterate on the react experience on main until we're happy with it, and then port the decisions to the other UI lib implementations.

travis:
+1 to all these suggestions, would love to get them landed asap!

alanshaw (approved):

PR #209, opened 2023-01-05
chore: migrate react examples to vite

travis:
following the example of https://github.com/web3-storage/w3ui/pull/203, move the rest of the react examples to vite

PR #210, opened 2023-01-05
chore: use `prepare` script to auto-build after install

olizilla:
Adds a [`prepare`](https://docs.npmjs.com/cli/v6/using-npm/scripts#life-cycle-scripts) script to automatically run `build` and `compile` after `install`. This gives us one fewer steps to get started and is a good habit as prepare also runs if folks depend on your modules from a git branch or similar.

see: https://docs.npmjs.com/cli/v6/using-npm/scripts#life-cycle-scripts

Also:
- updates an old ref to `npm` in the scripts
- removes `build` and `compile` from CI action as it's included in `install` now thanks to `prepare`
- updates example readmes to remove `build` step
- adds us as maintainers!
- adds `package-lock.json` to gitignore so we don't get double lock madness.

License: MIT


olizilla:
Odd that we also have to run `compile` too. I'm gonna add that to `prepare` as well.

olizilla:
This should wait for https://github.com/web3-storage/w3ui/pull/209 as we need that to get the react examples working. i'll deal with the merge conflicts after that one lands.

hugomrdias (approved):
nice lgtm

yusefnapora (approved):
👍

travis:
This looks good to me, though I will note that `build` is no longer strictly necessary in many cases because `main` now points directly at TypeScript files in development and all of the packages and examples know how to deal with TypeScript directly. I'm a little worried it's going to be annoying to have the overhead of a `build` every time we run `pnpm install` but I'm down to see how it goes!

travis (approved):

olizilla:
@travis fair point about being forced to run build after every `install`... perhaps if we are able to drop or reduce the amount of work build does then it won't be too annoying. Let's try it and I can revert if it's a pain.

Issue #156, opened 2022-12-10
proposal: add services-core and react-services packages to encapsulate upload/access service configuration


travis:
I started trying to write a `ServicesContext` that can optionally be configured to point w3ui components at staging, production or another branch. I realized this probably needs to be done in its own set of packages (one core, and one for each framework we support). 

I'll probably tweak some typing in this process (ie, split the two slightly different things currently called ServiceConfig into UploadServiceConfig and AccessServiceConfig) 

@alanshaw just want to make sure you don't have something in process around this

jchris:
Is it possible to hoist this into the w3up-client layer?

travis:
half of it yep! per discussion this morning with @gobengo @alanshaw @hugomrdias @Gozala it sounds like the package that actually encapsulates the service details should be outside this repository entirely so that other services and libraries can all depend on it as well.

we will still need to implement framework specific configuration components (ie, Providers in React, something else in other frameworks) and those should probably still be in their own package, worth leaving this issue open for that

gobengo:
IMO we may not really need this.
esp after https://github.com/web3-storage/w3protocol/issues/325#issue-1519480140
Insofar as the 'services' package would contain ucanto connection endpoints and ucan `aud` principal IDs that will respond to w3protocol caps, I think after that issue no one will even need those anymore.

They'll just need to know whether they are trying to send invocations to `did:web:web3.storage` or `did:web:staging.web3.storage`, and they can resolve those URIs using http, maybe follow a link or two, to get to a ucanto http endpoint (and the service principal is the `did:web` they already have).

travis:
oh that will definitely make this simpler - I thiiiink we'll still want a package that provides affordances for easily changing which service the w3ui components are using (ie, lets you override which DID the `{Upload|UploadsList|Keyring}Provider` tries to send invocations to) but I think it will be pretty basic and possibly not even needed once the proposal you linked is implemented.

PR #215, opened 2023-01-05
chore: add ava config and some basic tests

yusefnapora:
This needs some cleanup, but I got ava + typescript to work after a bit of wrestling. 

So far I've only added some simple tests to keyring-core that don't depend on the service, but I've also started on a mock access service so we can test `registerSpace`, etc.

I had to add `"type": "module"` to the keyring-core package.json to get the ava + typescript setup to work. That seems pretty safe, since we're only including the rollup bundle in the published package, so it shouldn't make a difference to users.

It seems like the mock services would be useful for all the packages, so maybe we should make a shared `test-utils` package?


packages/keyring-core/test/utils/mock-service.ts
```diff
@@ -0,0 +1,56 @@
...
+      return undefined
+    }),
+    recover: provide(Space.recover, async () => {
+      console.log('space/recover')
+      return []
+    })
+  }
+}
+
+export async function mockAcccessConnection (): Promise<ConnectionView<Service>> {
```
travis: extra "c" here

packages/keyring-core/test/agent.ts
```diff
@@ -0,0 +1,41 @@
+import test from 'ava'
+import { JSDOM } from 'jsdom'
+import 'fake-indexeddb/auto'
+
+import { createAgent } from '../src/index.js'
```
travis: we can get rid of these extensions if we use `[--experimental-specifier-resolution=node](https://nodejs.org/api/esm.html#customizing-esm-specifier-resolution-algorithm)` option mentioned in https://github.com/avajs/ava/blob/main/docs/recipes/typescript.md#for-packages-with-type-module - I saw this earlier this week poking around at AVA but didn't get as far as you have here and am delighted to see it works locally. I'll push a commit doing this, happy to back it out if we don't want to use this

travis (approved):
this is fantastic!! I left a couple comments but already pushed fixes for them - happy to back either of my changes out if you don't want them, but I was in the code testing the `--experimental-specifier-resolution` change and it was easy to knock them both out

travis:
this is already strictly better than what we have now so I'd be happy if you merged as-is!

> It seems like the mock services would be useful for all the packages, so maybe we should make a shared test-utils package?

I think it might be a little cleaner for each package to export useful test utils and to introduce a `services` package of some sort per https://github.com/web3-storage/w3ui/issues/156 where we can implement service Providers and land these mock services - I'm worried that a shared `test-utils` package will end up depending on all these packages, and that all the packages will end up dev-depending on `test-utils` leading to a gnarly knot of dependency resolution. Strictly speaking I think either will work, but the idea of a dependency knot like this puts a knot in my stomach...

yusefnapora:
awesome, thanks for the fixes @travis :)

I think you're right that the services package is where it's at. Should we just pull the WIP mock service out of this PR, so it's not cluttering up the joint? It's not being used yet (except in the skipped test).

travis:
>  Should we just pull the WIP mock service out of this PR

Yep that sounds great - probably time to create a new package (`@w3ui/services`?) where that can live. If you're up for creating it go for it, otherwise I'm happy to get that into a draft PR in the next day or so.

travis:
oh also maybe just singular `@w3ui/service` since @gobengo unified the two services into one in https://github.com/web3-storage/w3protocol/issues/325#issue-1519480140

yusefnapora:
sweet, this is passing on CI now that I bumped it up to node 18.

I'll go ahead and merge now that the mocks are gone. `@w3ui/service` sounds good to me for the new package. I can start on that tomorrow unless you feel like grabbing it sooner :)

PR #220, opened 2023-01-06
fix: make react sign-in example work

olizilla:
rename index.js to index.jsx so vite knows to expect jsx tags in the file.

also:
- updates react `template` app in the same way.

License: MIT


travis (approved):
thanks @olizilla !

alanshaw (approved):

Issue #150, opened 2022-12-07
add tests


travis:
UI testing! A very special hell that we should probably brave sooner than later.

We currently have a codesandbox integration configured that at least builds the example apps in CI. 

We should add tests similar to the ones in the HeadlessUI component library, eg:

https://github.com/tailwindlabs/headlessui/blob/main/packages/%40headlessui-react/src/components/combobox/combobox.test.tsx

Now that we have a bit of complexity in this library, some basic testing will probably make our lives easier down the road.

travis:
I spent a bit of time before the break on this and ran into some annoying issues where Jest had trouble loading ESM dependencies. @hugomrdias mentioned yesterday that he was interested in looking into alternatives to Jest, which I'd be excited about based on this experience and others, and spent a bit of time researching the state of the art during which I found this article:

https://whistlr.info/2022/test-react-builtin/

TLDR is 

a) Node 18 comes with a builtin testing library `node:test` that uses ESM rather than insisting on CJS as Jest does, which should help with the issues I was having
b) `node:test` is definitely much more barebones than Jest or Mocha, and needs some sugar to feel as slick as Jest or Mocha
c) because `node:test` uses ESM, mocking third party dependencies is hard:
<img width="800" alt="Screen Shot 2023-01-03 at 1 59 30 PM" src="https://user-images.githubusercontent.com/1113/210448358-e6de11c9-a451-4ab2-b67a-d05fd6cb86de.png">


Given all this I'm leaning slightly in the direction of trying to use `node:test` but probably need to discuss further with anyone interested. Would love to know what others think/if anyone has experience with this stuff and can provide an informed perspective!

gobengo:
@travis in general I've been told (by @alanshaw iirc) that we should use [`ava`](https://www.npmjs.com/package/ava) for tests. It's a good middle ground that supports ESM.

travis:
this has been broken down into #222 and #223. keeping this around to track the higher level intent to add more testing but we can close it once the sub-issues are closed.

olizilla:
here is a nice example of reusing the vite preview server for a browser based playwright test from vitest.
https://github.com/vitest-dev/vitest/blob/main/examples/playwright/test/basic.test.ts

for integration tests, i think re-using the vite preview server would be a win, and using playwright to handle browsers for us has been the least worst option we've hit so far.

[...]

PR #238, opened 2023-01-09
feat: test examples in playwright

olizilla:
- use playwright and playwright-test to verify the examples run in a browser
- have playwright-test fire up the vite preview server on a known port to serve the example
- add github workflow

So far i've just added a minimal e2e style test for the vue sign-in-up example, to show the pattern and get feedback. It's already found a bug there in safari which doesn't appear in firefox or chrome:

Screenshot 2023-01-09 at 14 30 39

Screenshot 2023-01-09 at 14 30 53

WIP on #150 

### TODO

- add tests for other examples
- consider pulling tests up to parent dir and reuse the same e2e test for each ui flavour - i'm imagining we'd have a test for each example, and then run that suite once for each ui library directory.
- fix safari bug in sign-in example where submitting the form errors and returns the user to the form instead of the verify message

License: MIT

### DIFF
--- a/.codesandbox/ci.json
+++ b/.codesandbox/ci.json
@@ -1,5 +1,4 @@
 {
-  "buildCommand": false,
   "sandboxes": [
     "/examples/react/file-upload",
     "/examples/react/sign-up-in",
--- a/.codesandbox/ci.json
+++ b/.codesandbox/ci.json
@@ -1,4 +1,5 @@
 {
+  "buildCommand": false,
   "sandboxes": [
     "/examples/react/file-upload",
     "/examples/react/sign-up-in",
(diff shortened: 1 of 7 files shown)


gobengo:
lookin good. @olizilla I noticed the PR description here says `playwright-test` ([npm](https://www.npmjs.com/package/playwright-test)) but the code uses `@playwright/test` ([npm](https://www.npmjs.com/package/@playwright/test)).

olizilla:
@gobengo yes! good point! it's confusing as heck, but in this instance I mean to use `@playwright/test`  for its comfy UI testing.

I made a better one over here #242 where we can write the tests for the examples once and then run them over all the ui libraries.

PR #239, opened 2023-01-09
chore: replace ava with vitest

yusefnapora:
Based on the discussion in #150, I decided to try replacing ava with vitest and found that it was super easy 😄 Much nicer out-of-the-box experience than getting ava to play nice with typescript.

Any objections to tossing ava overboard? edit: once the build failure is sorted out, that is 😐


yusefnapora:
ugh. I need a way to turn a PR back into a draft :)

This was working locally until I did a `pnpm clean` and reinstall... now I get the same error as on CI. Will poke some more...

yusefnapora:
The errors I was getting were due to the [webcrypto API's not being defined by jsdom](https://github.com/jsdom/jsdom/issues/1612). I have no real clue why it was working for me locally...

It seems like the way to customize the vitest environment is to define [a custom environment package](https://vitest.dev/guide/environment.html), so this adds a `vitest-environment-w3ui` package that just wraps the existing JSDOM package and sets the global `crypto` var to Node's `crypto.webcrypto` implementation.

I used this [miniflare vitest env](https://github.com/cloudflare/miniflare/blob/master/packages/vitest-environment-miniflare/src/index.ts#L78) as inspiration for updating the crypto global.

packages/keyring-core/test/agent.ts
```diff
@@ -1,29 +0,0 @@
-import test from 'ava'
-import { JSDOM } from 'jsdom'
-import 'fake-indexeddb/auto'
-
-import { createAgent } from '../src/index'
-
-test.before((t) => {
-  const dom = new JSDOM('<!DOCTYPE html>')
-  globalThis.document = dom.window.document
```
olizilla, line 9: how did this work with the `crypto` dependency previously?

olizilla (commented):
This looks good. I'm just trying to get me head around why the crypto global needs to be set explicitly for vitest but didn't for ava. My guess is that the code under test in ava had access to node crypto, while vitest would only give it's test code access to the globals configured in `jsdom` if configured with `environment: 'jsdom'`

packages/keyring-core/test/agent.ts
yusefnapora: I'm not sure what the difference was - it's especially confusing that it was working with my stale node_modules...

The error was coming from this bit in `@noble/ed25519`:

```js
    sha512: async (...messages) => {
        const message = concatBytes(...messages);
        if (crypto.web) {
            const buffer = await crypto.web.subtle.digest('SHA-512', message.buffer);
            return new Uint8Array(buffer);
        }
        else if (crypto.node) {
            return Uint8Array.from(crypto.node.createHash('sha512').update(message).digest());
        }
        else {
            throw new Error("The environment doesn't have sha512 function");
        }
    },
```

And the error message was that `digest` was undefined. Which is a bit odd, since it suggests that `crypto.web` was defined, but `crytpo.web.subtle.digest` was not... Best guess is that the vitest jsdom env explicitly sets the `crypto` var to some non-empty value that's not actually backed by the crypto API. But that seems like a weird thing to do, so maybe not...

yusefnapora: I added some console logs to the vitest plugin to try to figure out what's going on. Seems like they define a `crypto` global with just the `getRandomValues` function defined...

Here's my logging code:

```js
    console.log('!!! old crypto global:')
    console.dir(global.crypto)
    for (const p in global.crypto) {
      console.log(`crypto prop ${p}: ${global.crypto[p]}`)
    }
```

And the output:

```
!!! old crypto global:
Crypto {}
crypto prop getRandomValues: getRandomValues(array) {
      const esValue = this !== null && this !== undefined ? this : globalObject;
      if (!exports.is(esValue)) {
        throw new globalObject.TypeError(
          "'getRandomValues' called on an object that is not a valid instance of Crypto."
        );
      }

      if (arguments.length < 1) {
        throw new globalObject.TypeError(
          `Failed to execute 'getRandomValues' on 'Crypto': 1 argument required, but only ${arguments.length} present.`
        );
      }
      const args = [];
      {
        let curArg = arguments[0];
        if (ArrayBuffer.isView(curArg)) {
        } else {
          throw new globalObject.TypeError(
            "Failed to execute 'getRandomValues' on 'Crypto': parameter 1" + " is not of any supported type."
          );
        }
        args.push(curArg);
      }
      return esValue[implSymbol].getRandomValues(...args);
    }
```

yusefnapora: This seems to solve the mystery of why it worked with ava, although I still don't know why it was working with my stale node_modules. 

With ava, the `crypto` global was undefined, so `@noble/ed25519` took the "node crypto" branch. But with vitest we have just enough `crypto` to lead us down the webcrypto branch, but no `crypto.subtle` to seal the deal.

package.json
```diff
@@ -49,6 +51,8 @@
     "ts-node": "^10.9.1",
     "ts-standard": "^12.0.1",
     "typescript": "^4.8.2",
+    "vitest": "^0.27.0",
+    "vitest-environment-w3ui": "workspace:1.0.0",
```
travis, line 55: I think the version string here can just be "workspace:^"

rollup.config.ts
```diff
@@ -156,6 +156,16 @@ export default function rollup (options: RollupOptions): RollupOptions[] {
         vue: 'Vue',
         '@w3ui/vue-keyring': 'VueKeyring'
       }
+    }),
```
travis, line 159: do we need rollup to build this? afaik the rollup output is only used in built packages, and I'm not sure we want this particular package to be published. could be wrong, just curious what needs this...

travis (approved):
couple minor comments, most importantly addressing whether we need rollup, but happy to see this merge

olizilla:
🎉 🚀

PR #241, opened 2023-01-11
chore: add headless react components to markdown docs

yusefnapora:
This just extends #136 to add the new components to the markdown docs in the `docs/` dir. I pretty much just copied over the doc comments, so nothing really ground breaking here :) 

@travis, do you already have any simple example snippets I could crib? It would be nice to show a minimal example that renders each component. We can do that in a follow-up PR though if you want to just pull this in now.


travis (approved):
This looks great - we'll need to update later with details about the `as` property from https://github.com/web3-storage/w3ui/pull/236 but this is a great start.

The best place to find snippets is actually in the CustomizableUI components in https://github.com/web3-storage/w3ui/pull/208 - see, eg https://github.com/web3-storage/w3ui/pull/208/files#diff-60c8d15de54bb395efe3cb81c188d666d0bfecdd8c163500ed56335de986003bR50 for an example of how the Headless components can be used to build more complicated components

Issue #222, opened 2023-01-06
add integration tests


travis:
Chatted with @olizilla and @yusefnapora this morning and we are inclined to do browser testing with some third party service that makes it easy and low maintenance. Need to investigate the state of the art for this kind of testing.

olizilla:
There's "the test framework" and "the cloud test service", and some frameworks can be run by multiple services

| test framework | cloud service | feels like |
|----------------|---------------|----------|
| cypress             | [cypress cloud](https://docs.cypress.io/guides/cloud/introduction#Benefits), [browserstack](https://www.browserstack.com/automate/cypress), lambdatest | mocha + chai |
| playwright-test | [lambdaTest](https://www.lambdatest.com/playwright-testing) | ava + expect (but you can [drive `playwight` from any test framework](https://playwright.dev/docs/test-runners)

olizilla:
Added integration tests using @playwright/test to drive the browser orchestration in #242 as the first part of this. I think the scope of this issue should be "have integrations tests we are happy with". We can adopt a testing service if we find friction or features that we need.

PR #242, opened 2023-01-11
feat: test examples in real browsers

olizilla:
Pulls example tests out to own dir so we can run the same test for each ui library! 

fixes: #222 

You can see it run in CI at https://github.com/web3-storage/w3ui/actions/runs/3901558574/jobs/6663550379#step:6:13 and you can try it out locally from this branch by running:

```
$ pnpm install

# one time download of test browsers
$ pnpx playwright install

# fails for vanilla currently, but is ok, proceed.
$ pnpm build:examples 

$ pnpm test:examples

Scope: 20 of 33 workspace projects
examples/test/playwright test$ playwright test
[30 lines collapsed]
│ [30/33] [webkit] › sign-up-in.spec.ts:4:3 › vue: sign in
│ [31/33] [webkit] › uploads-list.spec.ts:4:3 › react: uploads list
│ [32/33] [webkit] › uploads-list.spec.ts:4:3 › solid: uploads list
│ [33/33] [webkit] › uploads-list.spec.ts:4:3 › vue: uploads list
│ 
│   33 passed (7s)
│ To open last HTML report run:
│ 
│   npx playwright show-report
│ 
└─ Done in 7.6s
```

_note: the build fails currently as our vanilla examples fail to build currently, which is why we need these tests!_

Adds and updates run scripts to make it easy to run package or example tests. Most of the time we'd want to run package tests locally, so this is set as the default.

```bash
# runs package tests
$ pnpm test

# builkd examples. must be run before test so the dist dir exits for each
$ pnpm build:examples

# run example test
$ pnpm test:examples

# serve examples, so you can poke around as a dev. 
$ pnpm serve:examples
```

Updates examples to produce relative URLs by setting `base: ''` in vite config. This allows us to serve all the examples from a single static server in the `examples` dir, and `serve` is added with a `serve.json` config to rewrite the paths so the `dist` dir is served for each example.

Adds playwright config to fire up the example server

Adds github workflow to set up and run the browser tests only if we make a change to the examples. We may change this to run for every change in the future, as we should probably check if the examples still work if we change a package!

Reduces the scope of our local github action to just handle setting up node and pnpm which is needed by every workflow. `lint` and `test` are now specified explicitly in workflows that should run them. We don't need to run those steps when we're doing browser testing.

- Vanilla JS examples are not currently tested as they fail to build https://github.com/web3-storage/w3ui/issues/243 - we can fix in a separate PR.
- We need a mechanism to be able to register a space from browser, or add pre-registered test account fixture - https://github.com/web3-storage/w3ui/issues/244
- We are missing a `multi-file-upload` example - https://github.com/web3-storage/w3ui/issues/245

Screenshot 2023-01-12 at 02 03 41

Screenshot 2023-01-12 at 01 57 39

**whoo!**
Screenshot 2023-01-12 at 11 00 28

License: MIT


olizilla:
I explored using vitest to drive playwright tests. It's ok, but a little clunky and I think we want as little friction as possible for browser testing. i'm now inclined to use @playwright/test instead after trying it in #238 and now i see it has support for what we need to do here https://playwright.dev/docs/test-advanced#projects and https://playwright.dev/docs/test-parameterize and if we stick with vanilla @playwright/test we can use a service to run our tests for us if we want to.

olizilla:
Here's a good example of the kind of info we get from this... Our vue `multi-file-upload` example is having a bad day. Passes in all browsers for other ui libraries

Screenshot 2023-01-12 at 11 17 29

examples/test/playwright/src/file-upload.spec.ts
```diff
@@ -0,0 +1,13 @@
+import { test, expect } from '@playwright/test'
+
+for (const ui of ['react', 'solid', 'vue'/* 'vanilla' not work atm */]) {
```
travis, line 3: this is dope af

travis (approved):
I love it! I'm going to start a discussion around which examples we want to keep longer-term so I wouldn't go too much deeper on this for now, but I'm super excited about the approach!

olizilla:
@travis thanks! i'm gonna land this but i'm more than happy to iterate on it to work with #256

PR #253, opened 2023-01-12
fix: headless authenticator bugfixes

travis:
1. add email to the dependency list for `handleRegisterSubmit`
2. load the agent once at load time to ensure we use existing sessions

### DIFF
--- a/packages/react-keyring/src/Authenticator.tsx
+++ b/packages/react-keyring/src/Authenticator.tsx
@@ -1,6 +1,8 @@
-import React, {
-  useState, createContext, useContext, useCallback, useMemo, useEffect
-} from 'react'
+import type { As, Component, Props, Options } from 'ariakit-react-utils'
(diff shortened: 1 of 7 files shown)


packages/react-keyring/src/Authenticator.tsx
```diff
@@ -49,6 +49,13 @@ export const AuthenticatorContext = createContext<AuthenticatorContextValue>([
   }
 ])
 
+export const AgentLoader = ({ children }: { children: JSX.Element }): JSX.Element => {
+  const [, { loadAgent }] = useKeyring()
+  // eslint-disable-next-line
+  useEffect(() => { loadAgent() }, []) // load agent - once.
```
olizilla, line 55: Where was `loadAgent` called from previously?

olizilla (approved):
LGTM.

travis:
this was merged into #236 and so shipped with that

PR #254, opened 2023-01-12
fix: automatically load the first page of results

### LINKED ISSUE #200: proposal: React useUploadsList hook should automatically load one page of results
while fixing the react uploads list example (https://github.com/web3-storage/w3ui/pull/195) I noticed that the useUploadsList hook doesn't automatically load the first page of results - I think this is a bug but figured I'd double check before changing the current behavior.

travis:
resolves #200


olizilla (approved):
👍

travis:
whoop - I forgot this was merged into #236, so this technically shipped with that change

Issue #225, opened 2023-01-06
create new w3console project


travis:
I think it makes the most sense to create this as an example project in w3ui based on the branch in #208 and then to move to its own repository once that branch lands

travis:
Big open question is: what framework do we use? Currently leaning toward vite + preact but open to other ideas. 

Advantages of vite + preact:

1) we're now using vite across projects and it provides a very nice harness for dev and testing
2) preact in React-compatibility mode will demonstrate compatibility with React and Preact 
3) this approach will make w3console more easily composable with existing apps, which isn't easy with something like Next.js

gobengo:
> preact in React-compatibility mode will demonstrate compatibility with React and Preact

React is what we use [@web3-storage/website](https://github.com/web3-storage/web3.storage/tree/main/packages/website) and [nft.storage/website](https://github.com/nftstorage/nft.storage/tree/main/packages/website) today. React has compatability with React. In general, I think it's fair to claim that the addressable audience of devs-who-have-used-react is much bigger than that of devs-who-have-used-preact. So personally I would recommend using 'just react' and not 'preact', or at least would point out that there are no benefits to preact-over-react presented here yet.

> this approach will make w3console more easily composable with existing apps, which isn't easy with something like Next.js

If composability is a goal, and I think it's a good one, I think a good way of guaranteeing it is to develop it as a component, but demo it inside a component gallery like [storybook](https://storybook.js.org/). If you're demoing something in storybook, it's guaranteed that it's composable. If not, whether the thing is using next.js or vite, it may still be not composable. You can do [storybook w/ vite](https://storybook.js.org/blog/first-class-vite-support-in-storybook/).

If you dev a `W3Console` component in storybook + vite, we maintain the most optionality for someone else wanting to come along and embed that component in some other project using whatever combination of next/vite/etc or react/preact etc.

travis:
> In general, I think it's fair to claim that the addressable audience of devs-who-have-used-react is much bigger than that of devs-who-have-used-preact.

Yep totally agree here - the idea here is that if we use Preact we'll essentially be targetting React and Preact developers at the same time since Preact provides a subset of React's functionality. That said I've never actually used Preact for anything serious, so I'm not 100% sure this is true - @olizilla @yusefnapora I don't remember if either of you has more experience with Preact, but curious about your take here.

> I think a good way of guaranteeing it is to develop it as a component, but demo it inside a component gallery like [storybook](https://storybook.js.org/).

[...]

PR #267, opened 2023-01-18
feat: add storybook

olizilla:
Add storybook scaffolding so we can iterate on our components and all the states they can be in.

Screenshot 2023-01-18 at 16 58 32

Uses the beta release of storybook v7 as it has support for pnpm.

Part of #225 - we need this to demonstrate out components in all their possible states so we can make them good.

License: MIT

### DIFF
--- a/examples/react/playground/.storybook/main.cjs
+++ b/examples/react/playground/.storybook/main.cjs
@@ -1,0 +1,19 @@
+const path = require('path');
+module.exports = {
+  "stories": [
+    "../src/**/*.mdx",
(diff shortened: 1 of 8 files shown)


olizilla:
Using `context` for all the properties of our components makes it harder to set them up in storybook. We might want to export basic UI components so we can just pass in state as props and make it easier the styling part easier to iterate on.

storybook expects us to statically configure Args that would be passed in as props to our components, but we have to use the escape hatch of the render fn, as demo'd in this PR to configure a context instead of passing in props.

travis (approved):

PR #284, opened 2023-01-19
fix: w3console polish

travis:
first up: improved space creation flow

There are few different ways this could work, but the fact that we store `currentSpace` in the Agent combined with the fact that that `currentSpace` is updated automatically when we call `createSpace` makes it harder to implement some designs.

Put another way, it would be hard to have a space creator that does not change the currently selected space, so these changes lean into the way state management works right now.

https://user-images.githubusercontent.com/1113/213601254-86fd1114-e6d3-45a6-add3-44100166904a.mp4

### DIFF
--- a/examples/react/w3console/package.json
+++ b/examples/react/w3console/package.json
@@ -10,6 +10,7 @@
     "preview": "vite preview"
   },
   "dependencies": {
+    "@heroicons/react": "^2.0.13",
     "@w3ui/keyring-core": "workspace:^2.0.1",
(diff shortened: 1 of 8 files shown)


travis:
added a commit with a revamped uploader flow per some ideas from @olizilla - much nicer!

https://user-images.githubusercontent.com/1113/213600129-e8119a92-5c26-4e22-ae58-41626654c102.mp4

olizilla (approved):
🚀 ✨

PR #288, opened 2023-01-20
fix: load first page of uploads list in provider

travis:
rather than doing this in the headless component, do it in the hook so that hook users also get a page of results. originally had this in the provider but that had timing issues and only worked sometimes.

### DIFF
--- a/packages/react-uploads-list/src/providers/UploadsList.tsx
+++ b/packages/react-uploads-list/src/providers/UploadsList.tsx
@@ -13,8 +13,8 @@ export const uploadsListContextDefaultValue: UploadsListContextValue = [
     loading: false
   },
   {
-    next: async () => { },
-    reload: async () => { }
+    next: async () => {},
+    reload: async () => {}
   }
 ]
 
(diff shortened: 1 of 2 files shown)


packages/react-uploads-list/src/providers/UploadsList.tsx
```diff
@@ -13,8 +13,8 @@ export const uploadsListContextDefaultValue: UploadsListContextValue = [
     loading: false
   },
   {
-    next: async () => {},
-    reload: async () => {}
+    next: async () => { },
+    reload: async () => { }
```
jchris, line 17: Should we isolate white space changes to a code base wide pr?

jchris (approved):
My comment is not a blocker, just maybe means we are due for a linter commit

travis:
just a heads up to reviewers that the churn between different solutions here is exactly the kind of thing that a nice test can help clarify and coincidentally I was just saying I want to write some tests before we cut a release - I'm gonna write some nice tests for this tonight or tomorrow

olizilla (approved):
I think this is what most use cases will want, so 👍 

if folks need flexibility to not load the first page on first load then they'll tell us!

travis:
lol, innocent me upthread thinking a test for this would be simple - spent a bunch of time working out how to do this and made some progress, but decided to merge as-is and come back later with tests - mocking the service interaction and dealing with async functions running in the provider is extra spicy testing fun - doable, but will take a bit more work than I'd like to do before getting this in

PR #289, opened 2023-01-21
    feat: speed run photo uploader

travis:
very ugly, but a demonstration of the simplest possible photo upload app, implemented in 5 minutes:

https://user-images.githubusercontent.com/1113/213824973-02b5a344-534d-46ff-9efd-37f40c4997a3.mp4


jchris:
it would be cool to refine this app to stay basically as simple but leverage styling etc where possible, and then put it in the repo as an example of good code, but also have a 5 min speed run building it. None of this is urgent if what we have here is enough to unblock @gobengo's research.

PR #290, opened 2023-01-23
wip: fake keyring playground

travis:
I wanted to prove to myself that postMessage would be a viable way to communicate between an app and the keyring site, so I created two example apps and got them to talk to eachother.

to try this out, in two different consoles run:

console a:
```
cd examples/react/fake-keyring
pnpm install
pnpm run dev
```

console b:

```
cd examples/react/fake-keyring
pnpm install
pnpm run dev
```

then head to `http://localhost:3000` and hit "grab the keys"

you should see output in the console confirming this cross-origin communication works as expected, with some TODOs that sketch out what the keyring might need to do to make this all work as expected.

big open question for me - does the UCAN need to be transmitted all the way back to the app domain or can we just store it on the server as proof that a user has granted a domain access to a particular set of capabilities?

this should probably never be merged - close it when we no longer need it handy


gobengo:
> big open question for me - does the UCAN need to be transmitted all the way back to the app domain or can we just store it on the server as proof that a user has granted a domain access to a particular set of capabilities?

It probably depends on context, but in general I think we want more the former than the latter. Give the end-user (or their user-agent aka web browser) the delegation so they are in control of it. Don't hide it from them by storing it on their server without them having a copy.

Issue #256, opened 2023-01-13
proposal: consolidate example apps


travis:
I chatted with @yusefnapora a bit earlier this week about the future of our example apps. We'd tentatively like to reduce the number of examples to ease the maintenance burden. #255 brings us to 7 example apps for React alone (one of which is just a template, but still requires maintenance). Two of them, `ui` and `w3console` overlap with the other examples, demonstrating most of the functionality found in the more single-purpose examples like `file-upload` and `signin-on`. This is likely to become a huge pain for us, especially since we'd like to reproduce the examples in Vue, SolidJS and Vanilla JS.

For React, Vue and SolidJS, at least, it may be simpler to have 2-3 example apps broken up along abstraction lines: 
1. `w3console` in #255 does a good job of demonstrating how to use the "Customizable" UI components (from #208). 
2. I believe @natevw is working on an example app that shows how to upload photos to w3up, which may be a good use-case for the Headless UI-style components from #136. 
3. It may also be a good idea to have a simple app that demonstrates how to use just the Providers from each of the component packages without even the Headless UI components, but we might also skip that since it's relatively easy to see how to do that in the Headless UI components for a developer who wants this.

We may want to do something slightly different for the "vanilla" JS examples, as building out something as complicated as w3console in vanilla JS seems fairly complicated, but maybe not?

Relatedly, @olizilla introduced integration testing that can be written once and run against all the examples in all the frameworks in https://github.com/web3-storage/w3ui/pull/242 - I'd like to cohere around a strategy for examples apps as soon as possible so we can keep pushing that work forward.

If this sounds like a good idea to everyone, I'd like to move forward with getting apps 1 and 2 above polished enough to write tests against them. w3console as an app may diverge from the example app, but we'd keep it around as a useful teaching tool. I can work with @jchris and @natevw on what app 2 would look like. Once these two apps are done in at least a basic form, we could delete the other apps (maybe keeping around template for new developers?) and write tests.

Thoughts?

natevw:
fwiw re. the "app 2" discussion, I haven't been pursuing that recently but it would basically be a new version of https://github.com/Codigo-Fuentes/selfie-app-demo

cf. https://blog.web3.storage/posts/selfie-app which is a [somewhat bitrotted?] walkthrough of tutorial going *from* the react `file-upload` demo and converting it step-by-step into basically a `selfie-app-demo` end result.

olizilla:
[...]

PR #292, opened 2023-01-24
chore: get codesandbox examples working again

### LINKED ISSUE #237: codesandbox examples build but error when run
We've got codesandox-ci running on our PRs and it seems to build fine, but trying to run the examples fails with an error, and no debugging info.

of note:
- it only supports building via `npm` which is probably the cause of our examples not being runnable.
- codesandbox only supports up to node v16.

It may not be possible to make our build work in `npm` now we rely on features of `pnpm`, so we may have to disable `pnpm` ci, as it's a confusing signal for us to get green check on PRs even when the examples fail to run.

example build: https://ci.codesandbox.io/status/web3-storage/w3ui/pr/220
example sandbox: https://codesandbox.io/s/w3ui-example-react-sign-up-in-z6obt1

olizilla:
- codesansbox ci now support node 18
- disable `buildCommand` as it is done after install by the prepare lifecycle script

see: https://codesandbox.io/docs/learn/sandboxes/ci#configuration-format

fixes #237

note: there is an issue to consolidate the examples in #256 but our current set of examples are linked to from beta.ui.web3.storage so if we can get them working again without too much pain I think we should.

License: MIT

### DIFF
--- a/.codesandbox/ci.json
+++ b/.codesandbox/ci.json
@@ -1,4 +1,5 @@
 {
+  "buildCommand": false,
   "sandboxes": [
     "/examples/react/file-upload",
     "/examples/react/sign-up-in",
@@ -11,5 +12,5 @@
     "/examples/vue/uploads-list"
   ],
   "packages": ["packages/**"],
-  "node": "16"
+  "node": "18"
 }
--- a/.codesandbox/ci.json
+++ b/.codesandbox/ci.json
@@ -1,5 +1,4 @@
 {
-  "installCommand": "codesandbox:install",
   "buildCommand": false,
   "sandboxes": [
     "/examples/react/file-upload",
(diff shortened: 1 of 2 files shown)


.codesandbox/ci.json
```diff
@@ -1,4 +1,6 @@
 {
+  "installCommand": "codesandbox:install",
```
olizilla, line 2: `installCommand` must be a run script from the package.json, hence this carbuncle.

olizilla:
Also looks like codesandbox has added support for pnpm, as it's automatically detecting it in CI, see:

```
+ git checkout -qf remotes/origin/pull/292 (in /tmp/10863740)
+ node -v (in /tmp/10863740)
v18.12.1
+ pnpm -v (in /tmp/10863740)
7.18.2
+ pnpm install (in /tmp/10863740)
Scope: all 37 workspace projects
Lockfile is up to date, resolution step is skipped
```

travis (approved):
wooo!

PR #294, opened 2023-01-26
feat: Storybook story improvements

travis:
1) Add SpaceFinder to Storybook

SpaceFinder

2) Add Authenticator to Storybook

Authenticator

3) Break Tailwind classes in SpaceFinder, Authenticator and UploadsList down to their raw CSS so that the  components work and look good in Storybook.

UploadsList

4) Extract SpaceCreator to a new component

SpaceCreator


examples/test/playwright/package.json
```diff
@@ -4,7 +4,7 @@
   "type": "module",
   "scripts": {
     "serve": "serve ../../",
-    "test": "playwright test",
+    "test": "playwright install && playwright test",
```
olizilla, line 7: This can be reverted now #298 landed

examples/react/playground/src/stories/AuthenticationSubmitted.stories.tsx
```diff
@@ -0,0 +1,20 @@
+import { AuthenticationSubmitted } from '@w3ui/react'
```
olizilla, line 1: musing: perhaps we should make the naming consistent here so that we have `ComponentName` and `ComponentNameStateName`
```suggestion
import { AuthenticatorSubmitted } from '@w3ui/react'
```
like
- `SpaceCreator` -> `SpaceCreatorCreating`

or perhaps we should export a "view(s) only" component, that takes a state prop that we can set from a single story per Component, so our docs pages show all the things... tho that would force us to make a component that takes all the needed state and handlers for each of it's states, so perhaps the current pattern is better for now.

examples/react/playground/src/stories/AuthenticationSubmitted.stories.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+export default {
+  title: 'w3ui/AuthenticationSubmitted',
+  component: AuthenticationSubmitted,
+  tags: ['autodocs'],
+  argTypes: {
+  },
+  parameters: {
+    backgrounds: {
+      default: 'dark'
```
olizilla, line 12: this is neat!

olizilla: ooh let's set it in the defaults rather than per story for now
https://storybook.js.org/docs/react/essentials/backgrounds#configuration

olizilla: while we're noodling with defaults, _no one wants to see times new roman_ lets set at least a nice default font via
https://storybook.js.org/docs/react/configure/story-rendering#adding-to-body

packages/react/src/SpaceFinder.tsx
```diff
@@ -24,18 +24,18 @@ export function SpaceFinder ({ spaces, selected, setSelected, className }: Space
       )
 
   return (
-    <div className={className}>
+    <div className={`${className} w3ui-space-finder-wrapper`}>
```
olizilla, line 27: maybe use the no suffix class name for the root element like
```suggestion
    <div className={`${className} w3ui-space-finder`}>
```

olizilla: which is not normally a relevant metric, but if we want devs to feel nice when they customise our thingers it becomes more important!

examples/react/playground/src/stories/SpaceCreatorCreating.stories.tsx
```diff
@@ -0,0 +1,20 @@
...
+    backgrounds: {
+      default: 'dark'
+    }
+  }
+}
+
+export const Primary = {
+  args: {
+  }
+}
```
olizilla, line 20: same here, we can fix up the titles and exports to group states of the same logical components
```suggestion
export default {
  title: 'w3ui/SpaceCreator',
  component: SpaceCreatorCreating,
  tags: ['autodocs'],
  argTypes: {
  },
  parameters: {
    backgrounds: {
      default: 'dark'
    }
  }
}

export const Creating = {}
```

packages/react/src/styles/uploads-list.css
```diff
@@ -0,0 +1,64 @@
...
+}
+
+.w3-uploads-list nav {
+  display: flex;
+  flex-direction: row;
+  margin-top: 1rem;
+  margin-bottom: 1rem;
+}
+
+.w3-uploads-list nav>*+*{
```
olizilla, line 62: i have never seen one like that before!

olizilla (approved):
I love it. Some naming suggestions and tweaks suggested inline

examples/react/playground/src/stories/AuthenticationSubmitted.stories.tsx
travis: yeaaaaa - I'm not totally happy with the naming/structure here either, but similarly uncertain on the best "API design" here - I'll leave it as it is, but am doing a pass getting ready for Adam to come in and do some design work and will keep this in mind as I do, thanks!

examples/react/playground/src/stories/AuthenticationSubmitted.stories.tsx
travis: so real. will do!

packages/react/src/SpaceFinder.tsx
travis: agreed! I think there was a case in another component where I felt like `-wrapper` made sense, but this ain't it

packages/react/src/UploadsList.tsx
```diff
@@ -7,7 +7,7 @@ function Uploads ({ uploads }: { uploads?: UploadListResult[] }): JSX.Element {
     return (
       <>
```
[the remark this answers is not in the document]
travis: oh - we actually do this already in the `UploadsList` function at the bottom of this file - this function is private and called from that one, mainly to alias `props.uploadsList?.[0].data` to `uploads` without breaking the render-props style here

packages/react/src/styles/uploads-list.css
travis: right? cribbed from the tailwind docs! eg, https://tailwindcss.com/docs/space

examples/react/playground/src/stories/AuthenticationSubmitted.stories.tsx
```diff
@@ -0,0 +1,20 @@
+import { AuthenticationSubmitted } from '@w3ui/react'
+import '@w3ui/react/src/styles/authenticator.css'
+
+export default {
+  title: 'w3ui/AuthenticationSubmitted',
```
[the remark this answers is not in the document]
travis: SICK, great idea, this is much nicer

examples/react/playground/src/stories/SpaceCreatorCreating.stories.tsx
travis: ditto

Issue #297, opened 2023-01-27
playwright installing browsers at newer versions that the tests expect.


olizilla:
browser tests are failing in CI due to a mismatch between the browser versions installed and the versions depended on by tests.

```
Run pnpx playwright install --with-deps
...
[22](https://github.com/web3-storage/w3ui/actions/runs/4023317345/jobs/6914037373#step:5:23)
.../node_modules/playwright install: Chromium 110.0.5481.38 (playwright build v1045) downloaded to /home/runner/.cache/ms-playwright/chromium-1045
```

```
> pnpm run --filter './examples/**' test
...
[16](https://github.com/web3-storage/w3ui/actions/runs/4023317345/jobs/6914037373#step:6:17)
examples/test/playwright test:     browserType.launch: Executable doesn't exist at /home/runner/.cache/ms-playwright/chromium-1041/chrome-linux/chrome
```

see: https://github.com/web3-storage/w3ui/actions/runs/4023317345/jobs/6914037373

PR #298, opened 2023-01-27
fix: install correct browser versions for playwright tests

olizilla:
by running pnpm exec playwright install from the playwright dir, pnpm knows to use the version of playwright that we depend on rather than the latest version.

without this change we end up installing the latest browesers and then the tests fail as they can't find the older versions that they assume will have been installed.

...and it turns out `pnpx` is deprecated, and we should be using `pnpm exec` anyway.

fixes: https://github.com/web3-storage/w3ui/issues/297

see: https://pnpm.io/6.x/cli/exec

License: MIT

### DIFF
--- a/.github/workflows/examples.yaml
+++ b/.github/workflows/examples.yaml
@@ -18,7 +18,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: ./.github/actions/pnpm
       - run: pnpm build:examples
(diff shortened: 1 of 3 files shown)


olizilla:
This is a small PR and the goal is to fix the examples test... and the checks show that is done, so i'm gonna merge!

PR #308, opened 2023-01-28
feat: allow users to set page size in W3APIProvider

travis:
Allow users to set page size in W3APIProvider.

It feels like it would be cleaner to make it possible to set page size in the `useUploadsList` hook, but a bit of work on that revealed some complexity that will take a bit more planning and design to get right - see https://github.com/web3-storage/w3ui/commit/e50e02594bdf608b9add0241f22bed09c7dc2902 for the work I did toward this and a short description of the issues I ran into.

For now I think this is good enough to ship, and I'd like to think more carefully about how we handle state in the UploadsList in future versions of these libs.


examples/react/w3console/src/app.tsx
```diff
@@ -201,7 +201,7 @@ export function Logo (): JSX.Element {
 
 export function App (): JSX.Element {
   return (
-    <W3APIProvider>
+    <W3APIProvider uploadsListPageSize={10}>
```
olizilla, line 204: moar! i think we'll add fake multipage pagination to our list component like certain other consoles do.
```suggestion
    <W3APIProvider uploadsListPageSize={60}>
```

olizilla: so like "fetch 3 pages of 20, show the first 20 as "page 1" and then show "page 2" and "page 3" as links and then "..." to imply more pages, and then fetch the next n when the user flips through the pages, kinda thing

PR #309, opened 2023-01-30
feat: import a space into w3console

olizilla:
Add an import flow to add a space from a user provided ucan that delegates * capabilites on a space created by another agent.

Provides the w3console agent DID, so you can send it to the other party to create a delegation _just for you_

https://user-images.githubusercontent.com/58871/215516054-de617762-2e2d-4ff1-b6f9-c20772b46327.mov

License: MIT


olizilla:
I need to appease tsc on this one!

examples/react/w3console/src/share.tsx
```diff
@@ -22,15 +25,26 @@ export async function toCarBlob (delegation: Delegation): Promise<Blob> {
...
   const car = new Blob(carParts, {
-    type: 'application/vnd.ipld.car'
+    type: 'application/vnd.ipfs.car'
   })
   return car
 }
 
-export function SpaceShare (): JSX.Element {
-  const [, { createDelegation }] = useKeyring()
+export async function toDelegation (car: Blob): Promise<Delegation> {
```
travis, line 33: is there anything we should do here to ensure this is actually a delegation or is that handled in the library?

examples/react/w3console/src/share.tsx
```diff
@@ -80,9 +112,43 @@ export function SpaceShare (): JSX.Element {
...
+        {proof !== undefined && (
+          <div className='mt-4 pt-4'>
+            <Header>Added</Header>
+            <div className='max-w-3xl border border-gray-700 shadow-xl'>
+              {proof.capabilities.map((cap, i) => (
+                <figure className='p-4 flex flex-row items-start gap-2' key={i}>
+                  <DidIcon did={cap.with} />
+                  <figcaption className='grow'>
+                    <a href='#' onClick={() => viewSpace(cap.with)} className='block'>
+                      <span className='block text-xl font-semibold leading-5 mb-1'>{proof.facts.at(i)?.space.name ?? 'Untitled Space'}</span>
```
travis, line 139: I'd split this long line
```suggestion
                      <span className='block text-xl font-semibold leading-5 mb-1'>
                        {proof.facts.at(i)?.space.name ?? 'Untitled Space'}
                      </span>
```

examples/react/w3console/src/share.tsx
```diff
@@ -80,9 +112,43 @@ export function SpaceShare (): JSX.Element {
...
+                      <span className='block font-mono text-xs text-gray-500 truncate'>{cap.with}</span>
+                    </a>
+                  </figcaption>
+                  <div>
+                    <a href='#' className='font-sm font-semibold' style={{ verticalAlign: '-8px' }} onClick={() => viewSpace(cap.with)}>View</a>
```
travis, line 144: this one too, tho I'm not wedded to the division of props here so feel free to go a different way

also better to go with `align-[-8px]` rather than setting style manually: https://tailwindcss.com/docs/vertical-align#using-custom-values
```suggestion
                    <a href='#' className='font-sm font-semibold align-[-8px]' 
                       onClick={() => viewSpace(cap.with)}>
                      View
                    </a>
```

examples/react/w3console/src/share.tsx
```diff
@@ -80,9 +112,43 @@ export function SpaceShare (): JSX.Element {
             value={value}
             onChange={onChange}
           />
-          <a className='w3ui-button text-center block w-52 opacity-30' style={{ opacity: downloadUrl !== '' ? '1' : '0.2' }} href={downloadUrl ?? ''} download={downloadName(downloadUrl !== '', value)}>Download UCAN</a>
+          <a className='w3ui-button text-center block w-52' style={{ opacity: downloadUrl !== '' ? '1' : '0.2' }} href={downloadUrl ?? ''} download={downloadName(downloadUrl !== '', value)}>Download UCAN</a>
```
travis, line 115: similarly, I'd move this `style=` into the `className`
```suggestion
          <a className={`w3ui-button text-center block w-52 ${downloadUrl !== '' ? 'opacity-100' : 'opacity-20'}`} 
             href={downloadUrl ?? ''} download={downloadName(downloadUrl !== '', value)}>
            Download UCAN
          </a>
```

travis (approved):
This is amazing! Looks good, some minor style nits and questions but feel free to merge once you've placated `tsc`.

examples/react/w3console/src/share.tsx
olizilla: I had an issue with postcss not being able to detect dynamic classNames, which meant the class name wasn't included in the compiled css.

examples/react/w3console/src/share.tsx
olizilla: yes, the code should catch where the blob cannot be decoded as a CAR and let the user know that the file isn't a CAR (or corrupted). 

and it should catch an error from importDAG if the CAR doesn't contain blocks that can be parsed as a ucan delegation.

This feature in general is very much a proof of concept, so i'd prefer to land it as is, and then prioritise adding the good feedback if we decide it's a thing we want to keep. It'll be much less relevant once we can fetch users other spaces for them from accounts, and share spaces by email address.

olizilla:
The solid upload-list example integration test is failing. Am digging in.

PR #319, opened 2023-02-01
chore: upgrade all the deps

Gozala:
Updates all the deps and aligns versions so we don't have various version of the multiformat and @ipld/car

Most updates have been generate with a help of `pnpm upgrade --interactive -r  ` command, however for some reason it was failing to update deps in `examples` directory resulting in multiple versions of `multiformats` and `@ipld/car`. I end up working around this by manually changing `package.json`-s of packages that were referencing old versions.


examples/react/playground/package.json
```diff
@@ -32,10 +32,10 @@
     "@storybook/testing-library": "^0.0.13",
     "@types/react": "^18.0.26",
     "@types/react-dom": "^18.0.9",
-    "@ucanto/interface": "^4.0.3",
+    "@ucanto/interface": "^4.2.3",
```
Gozala, line 35: I think it would be a good idea for `@w3ui/react`  and perhaps others from the family to re-export same basics like `@ucanto/interaface` and perhaps some things from `multiformats` so that they don't need to be imported by the users and deal with dependency problems.

examples/react/playground/src/stories/UploadsList.stories.tsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react'
-import { CID } from 'multiformats/cid'
+import * as Link from 'multiformats/link'
```
Gozala, line 2: Note: CID still exists but `Link` is designed to be more abstract and less prone to becoming out of date.

travis (approved):
looks good to me!

PR #320, opened 2023-02-01
fix: fix w3console styling

travis:
I broke the w3console styling by failing to include some of the new stylesheets from the react components. Fix that, and introduce an omnibus CSS file called `all.css` to make this easier in the future.

Also tweak the Authenticator component to accept a top-level CSS class - I didn't end up using this, but implemented it while debugging and I think it's good to leave in.


jchris (approved):
this works great for me

PR #321, opened 2023-02-01
fix: add basic documentation for storybook and w3console

travis:
we need more in here eventually, but this gets us started


jchris (approved):
Thanks for document this

PR #347, opened 2023-02-02
chore: add tests for react-* headless components

travis:
Basic testing to verify the components in `Authenticator`, `Uploader`, and `UploadsList` call the expected API functions from their respective providers.


packages/react-keyring/test/Authenticator.spec.tsx
```diff
@@ -0,0 +1,88 @@
...
+test('CancelButton', async () => {
+  const user = userEvent.setup()
+
+  const cancelRegisterSpace = vi.fn()
+  const contextValue: KeyringContextValue = [
+    keyringContextDefaultValue[0],
+    { ...keyringContextDefaultValue[1], cancelRegisterSpace }
+  ]
+  act(() => {
+    render((
```
olizilla, line 34: I'm new to testing-library, but i like the look of it and its goals. From looking at some examples it looks like we should be able to avoid the beforeEach / afterEach boilerplate and maybe the `act` step can go too? If we can get our tests looking more like this example that'd be rad https://github.com/kentcdodds/react-testing-library-course/blob/5abc1039d677767b8de5197c049bfd0428fa7420/src/__tests__/app-03.js#L12-L38

olizilla: a handy list of things to avoid https://kentcdodds.com/blog/common-mistakes-with-react-testing-library which can be used to extrapolate what should be possible

olizilla (commented):
I think it's a good direction! Is it possible to reduce the boilerplate?

travis:
added a test for the Uploader as well

travis:
annnnd UploadsList

olizilla (approved):
This is great!

We should do some learning around getByRole and use our tests to check that our components follow aria guidelines. It's not a replacement for a11y testing, but it's a baseline we can put in early to ensure we get the basics. No need to change anything in this PR, I'm just surfacing a thought that occurred as I read through.

Issue #351, opened 2023-02-02
React UI Authenticator container styles impact on child layouts even when logged in.


olizilla:
We're applying a set of style to `.w3ui-authenticator` to allow us to show a full screen "not logged in" contsainer. 

https://github.com/web3-storage/w3ui/blob/0c661c6f4d28776b4ac7ad1aea3e0958186f8480/packages/react/src/styles/authenticator.css#L22-L28

That root class  remains in the tree even when you are logged in, which forces the applications layout to deal with being a flex child, and contained in a height:100vh parent element. 

The logged in UI should not have to deal with layout choices of the authenticator log in screen. Related to #349

PR #352, opened 2023-02-02
fix: remove authenticator class when registed

olizilla:
only include the `w3ui-authenticator` class in the dom when we want to show the log in box.

otherwise the layout we choose for the register form is inherited by the logged in app layout which is undesirable

fixes https://github.com/web3-storage/w3ui/issues/351

License: MIT

### DIFF
--- a/packages/react/src/Authenticator.tsx
+++ b/packages/react/src/Authenticator.tsx
@@ -5,13 +5,15 @@ export function AuthenticationForm (): JSX.Element {
   const [{ submitted }] = useAuthenticator()
 
   return (
-    <AuthCore.Form className='w3ui-authenticator-form'>
-      <div className='email-field'>
-        <label htmlFor='w3ui-authenticator-email'>Email address:</label>
-        <AuthCore.EmailInput id='w3ui-authenticator-email' required />
-      </div>
-      <button className='register w3ui-button' type='submit' disabled={submitted}>Register</button>
-    </AuthCore.Form>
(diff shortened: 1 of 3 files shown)


packages/react/src/Authenticator.tsx
```diff
@@ -48,7 +52,7 @@ interface AuthenticatorProps {
 
 export function Authenticator ({ children, className = '' }: AuthenticatorProps): JSX.Element {
   return (
-    <AuthCore as='div' className={`w3ui-authenticator ${className}`}>
+    <AuthCore as='div' className={className}>
```
jchris, line 55: So `w3ui-authenticator` means "needs authentication"? Maybe we want to rename to make clearer. Not a blocker for this PR.

jchris (approved):
runs correct for me on w3console

jchris:
Is there another test flow I should be trying out?

travis (approved):

packages/react/src/Authenticator.tsx
olizilla: kinda yes. From the pattern of our components it means "the root element for this widget", and this change makes it so the authenticator doesn't affect the dom if the user has > 0 registered spaces, so it implicitly takes on that meaning. 

I'm not making bigger changes here as we are all in favour of doing less of the "you must give an email to play", but if we revisit this component we'd probably add explicit css classes for "registered" and "not-registered" or similar.

PR #360, opened 2023-02-02
chore: document getCurrentSpace and getSpaces

travis:
Mainly doing this to get release-please to queue up a new keyring-core release after the previous one failed in a way that required updates to the build configs.


travis:
self-merging this because it's just a docs change and primarily here for its side effects

PR #368, opened 2023-02-03
fix: add more docs

travis:
`release-please` doesn't create a release for chores, so my last attempt to create a new keyring-core release failed.

Add some basic docs, and call this a `fix` to get `release-please` to do the right thing.


travis:
ok one more try at getting release-please to create a release for keyring-core - @yusefnapora I feel like my description of a Space leaves something to be desired, would love your thoughts on the best way to explain spaces to w3ui users and happy to update this ASAP.

for now, self-merging to get the release process going.

Issue #296, opened 2023-01-26
Fix linting and typechecking to match how we do in w3protocol.


olizilla:
ts-standard is not standard... it's significantly more opinionated, and a bit of a drain to work with. 

Out of numerous discussions on the subject over there years we eventually agreed to adopt the @hugomrdias method for (es)linting, prettiering and typechecking as demonstrated in w3protocol. this repo was started before that agreement, but we should take a moment to bring it inline with the current norms, lest it be just another repos with it's own way of doing things.

PR #380, opened 2023-02-07
chore: update linter, prettier configs

travis:
Per #296, update linter configs and install prettier to match https://github.com/web3-storage/w3protocol

Use @hugomrdias's `hd-scripts` configs, but turn off a few eslint rules that either a) I didn't like (no-negated-conditions, no-null) or b) feel duplicative in TypeScript (all of the jsdoc rules). Also turn off `import/extensions` because as far as I can tell it recommends the wrong thing. Without disabling these rules I got errors like this:

Screenshot 2023-02-07 at 4 32 03 PM

I'm definitely open to leaving some of these on if anyone likes them.

The first commit in this PR contains the config changes, the second is just the results of running `pnpm -w format` - it may be easier to review them separately to understand the changes.

Finally, I've only added eslint config in the `keyring-core` package for now - I believe I'll need to copy most of these changes into each of the packages, though would like to find a way to dry things up a bit - it may be possible to keep most of the config in a top level eslint config that each of the packages can inherit, and I'll keep investigating that while this is up for review and feedback.


travis:
Whoops - worth noting that prettier and the linter are currently in conflict - I will mediate.

olizilla (approved):
it does the thing! I wince at every change that deviates from that plain old `standard` does, but that is my own problem for another day. This brings us inline with w3protocol so I say YES.

olizilla (approved):
🚢 ✨

PR #383, opened 2023-02-20
chore: eject @w3ui/react from w3console

travis:
We don't have the bandwidth to support @w3ui/react anymore, so migrate w3console to use the "headless" components directly.

Importantly, this means we can use idiomatic TailwindCSS across the board rather than the hybrid raw-CSS/Tailwind approach we had been using.

A few years ago I would have considered the CSS changes here a step backwards, moving from semantic classes  to Tailwind utility classes sprinkled everywhere. While there may be some opportunities for CSS-class-based abstraction in a few places, Tailwind's docs make the case for being extremely conservative with this type of componentization:

https://tailwindcss.com/docs/reusing-styles

I have created a `w3ui-button` class, but it probably needs to be pared down - evidence for this can be found in the authenticator's "Register" button in which I needed to copy most of the styles from the `w3ui-button` class to customize it the way I wanted. I also created an `authenticator` class, but it may be a better idea to create a React component for this.

I have this configured to merge to `feat/pre-paging` for now to make this PR cleaner, but won't merge it until that hits `main` and this gets automatically retargetted.


examples/react/w3console/package.json
```diff
@@ -10,13 +10,15 @@
...
   },
   "dependencies": {
+    "@headlessui/react": "^1.7.7",
     "@heroicons/react": "^2.0.13",
     "@ipld/car": "^5.1.0",
     "@ipld/dag-ucan": "^3.2.0",
     "@w3ui/keyring-core": "workspace:^",
-    "@w3ui/react": "workspace:^",
     "@w3ui/react-keyring": "workspace:^",
+    "@w3ui/react-uploader": "workspace:^3.1.0",
```
alanshaw, line 19:
```suggestion
    "@w3ui/react-uploader": "workspace:^",
```

examples/react/w3console/package.json
```diff
@@ -10,13 +10,15 @@
...
     "@w3ui/react-uploads-list": "workspace:^",
+    "@w3ui/uploader-core": "workspace:^4.0.0",
```
alanshaw, line 21:
```suggestion
    "@w3ui/uploader-core": "workspace:^",
```

alanshaw (approved):

PR #384, opened 2023-02-21
fix: update to latest @web3-storage/access

travis:
this integrates the fix to https://github.com/web3-storage/w3protocol/issues/344 and enables working against staging services from this library


alanshaw (approved):

travis:
Updated 3 more deps. Going to merge based on the previous approval because they're substantially the same change, but happy to make updates after merge if there are any comments on the additions.

PR #400, opened 2023-03-15
feat!: use new account model

travis:
initial work on integrating new account model

* add new `authorize` function to the keyring context


docs/keyring-core.md
```diff
@@ -88,7 +88,7 @@ export interface KeyringContextActions {
    * storage. Use cancelRegisterSpace to abort. Automatically sets the
    * newly registered space as the current space.
    */
-  registerSpace: (email: string) => Promise<void>
```
gobengo, line 91: we may still want to allow passing a parameter here to indicate which account you want to use to add the storage provider to the space

gobengo: (e.g. you could have claimed delegations authorizing your issuer `did:key` to act as more than one account)

travis: Yep - it's not documented here but `registerSpace` also takes a map of options and I think that would be a better place for an email parameter if we do end up adding it to the lower level API - since the default case for this API is to just use the account we most recently `authorized` against I'd like to optimize for that both in the implementation and documentation for now.

Do you have any thoughts on user stories that might use the functionality you're proposing? I think in most cases where we'd be acting for more than one account we'd fire up two `Agent`s and authorize with them both separately (I'm imagining an "account selector" kind of interface).

examples/react/w3console/src/components/SpaceCreator.tsx
```diff
@@ -13,91 +13,100 @@ export function SpaceCreatorCreating (): JSX.Element {
...
+        /* eslint-disable no-console */
+        console.error(error)
+        /* eslint-enable no-console */
+        throw new Error('failed to register', { cause: error })
+      } finally {
+        resetForm()
+        setSubmitted(false)
+      }
+    } else {
+      throw new Error('cannot create space, no account found, are you logged in?')
```
alanshaw, line 50:
```suggestion
      throw new Error('cannot create space, no account found, have you authorized your email?')
```

alanshaw (approved):

olizilla:
You can set build time env for w3console here https://github.com/web3-storage/w3ui/blob/26e2af1e52b7082b65ac4b6c98c0ba804c0ef2eb/.github/workflows/w3console.yaml#L23-L25

if the build process uses them, values set there will be baked in, just before the dist is published to CF pages & w3

packages/keyring-core/src/index.ts
```diff
@@ -87,6 +87,10 @@ export interface KeyringContextState {
    * The current user agent (this device).
    */
   agent?: Signer
+  /**
+   * The account this device is authorized to act as. Currently just an email address.
+   */
+  account?: string
```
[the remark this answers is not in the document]
travis: ah, I like this - I'm going to file a separate issue but I want to jump on this today

PR #417, opened 2023-03-22
feat: add terms of service page

travis:
This required adding react-router-dom so we can have more than one page.

This definitely got me wondering if we shouldn't just move over to Next or Remix or someting that handles this and a lot more, but that probably requires a larger discussion. We should also discuss the middle ground of https://vite-plugin-ssr.com/ while we're at it.


.github/workflows/release.yml
```diff
@@ -92,6 +94,7 @@ jobs:
         working-directory: examples/react/w3console
         env:
           VITE_W3UP_PROVIDER: did:web:nft.storage
+          VITE_W3UP_SERVICE_BRAND_NAME: 'NFT.storage'
```
alanshaw, line 97: I think typeset like this normally:

```suggestion
          VITE_W3UP_SERVICE_BRAND_NAME: 'NFT.Storage'
```

alanshaw (approved):

PR #418, opened 2023-03-22
fix: add _headers

alanshaw:
IDK why but cloudflare pages is not setting `Content-Type` header. This PR adds a [`_headers` file](https://developers.cloudflare.com/pages/platform/headers) that sets them for the file types we currently generate.


olizilla (approved):
i am sorry it has come to this

PR #440, opened 2023-03-23
fix: tweak READMEs for packages that release-please didn't release right

travis:


travis:
self-merging to get the build going, happy to update after merge with any feedback

PR #463, opened 2023-03-24
fix: wait for register space before returning from space create onSubmit

travis:
because we weren't `await`ing here the space creation UI would flash empty again, which was very confusing


gobengo (approved):
those pesky forgotten awaits :/

consider adopting https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-misused-promises.md

PR #466, opened 2023-03-24
fix: get space registrar looking and acting right

travis:
it was still talking about email verification for registering spaces, and asking for an email. update it with all the latest good stuff


alanshaw (approved):
🚀

PR #467, opened 2023-03-24
feat: adjust language in register button

dchoi27:


alanshaw:
We're not really authenticating anything, we're _authorizing_ the local agent to act as the email account. I will change to "Authorize", but please shout if you think that's problematic and we can adjust.

alanshaw (approved):

dchoi27:
i guess the first time it's an authentication, and then future it's an authentication+authorization. but since we say register / authorization , think it's fine!

PR #471, opened 2023-03-24
fix: w3console email check page improvements

olizilla:
the colors match the login box, so are less jarring, and the cancel button is hidden! it doesn't work yet! See next PR.

**after**
Screenshot 2023-03-24 at 14 41 25

**before**
Screenshot 2023-03-24 at 14 41 50

License: MIT


travis (commented):
woo tyty!

PR #473, opened 2023-03-24
fix: wire up cancel registration button

olizilla:
but it fails to work after 2s. I think it's because once the ws connects we are stuck waiting for an ACK from the server.

by this i mean, you can click cancel, and it works as expected, taking you back the email input form... but only for ~2s... after which nothing happens for 60s, then the ws times out, and we are kicked back to the email form anyway.

License: MIT


travis:
I think this will need to change to `cancelAuthorize` after https://github.com/web3-storage/w3ui/pull/474 is merged, but the abort may work better! I think the core of this change - exporting a slightly different `cancelAuthorize` from `react-keyring` that sets `submitted` state - is still needed, I'm happy to update this PR once #474 is merged

alanshaw:
I think this is stale and was (re)implemented in https://github.com/web3-storage/w3ui/pull/591/commits/5bd52a87d6a0b08850edfcf583a6f4a1861e6b39

PR #474, opened 2023-03-24
fix!: update examples for new authorize flow

travis:
Unfortunately this also required some changes to the "keyring" interfaces - shipping another major release based on these changes is slightly annoying but probably the right thing to do, and honestly I'm probably the only person who will need to upgrade deps for now.

### DIFF
--- a/docs/keyring-core.md
+++ b/docs/keyring-core.md
@@ -84,15 +84,14 @@ export interface KeyringContextActions {
    */
   setCurrentSpace: (did: DID) => Promise<void>
   /**
-   * Register the current space, verify the email address and store in secure
-   * storage. Use cancelRegisterSpace to abort. Automatically sets the
-   * newly registered space as the current space.
+   * Register the current space and store in secure storage. Automatically
+   * sets the newly registered space as the current space.
    */
   registerSpace: (email: string) => Promise<void>
   /**
    * Abort an ongoing account registration.
    */
-  cancelRegisterSpace: () => void,
+  cancelAuthorize: () => void,
   /**
    * Get all the proofs matching the capabilities. Proofs are delegations with
    * an audience matching the agent DID.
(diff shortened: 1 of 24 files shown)


docs/keyring-core.md
```diff
@@ -84,15 +84,14 @@ export interface KeyringContextActions {
...
-   * newly registered space as the current space.
+   * Register the current space and store in secure storage. Automatically
+   * sets the newly registered space as the current space.
    */
   registerSpace: (email: string) => Promise<void>
   /**
    * Abort an ongoing account registration.
    */
-  cancelRegisterSpace: () => void,
+  cancelAuthorize: () => void,
```
gobengo, line 94: This would be backward incompatible, but the PR title does not have a `feat!`

travis: yeaaaa I think it was a bug in this major release that this wasn't changed, so I'd like it to ship in a minor release. I know it's technically semver heresy, but I also think it was just a bug that this didn't break with the last major

travis: put another way - if someone has already gone through the last major version upgrade but is depending on this function, I'd rather have their code break when they update the minor rather than requiring them to do another major version upgrade here, so my preference is to ship this in a minor

travis: Would still be helpful to get @alanshaw's brain on this one - @gobengo and I went back and forth but ultimately decided just shipping a new major release is probably the least ugly thing to do here, rather than shipping a minor release with a backwards incompatible change.

Probably a moot point given the current level of outside adoption of these libraries, but worth being on the same page about regardless.

olizilla (approved):
I think it's fine to release as major. 

More generally (not for this PR), i know that authorizing is now the step that can take a while (let's fix that too!) but I think all our methods that do a fetch should be cancellable in some way... like we should allow callers to pass in an `AbortSignal` so they can give up if their UI wants to.... any call to our api could stall.

PR #499, opened 2023-04-05
feat: upload progress

travis:
Use the new `onUploadProgress` callback in `@web3-storage/upload-client` to add upload progress bars to w3console and the example apps.

This PR also includes a decent amount of bugfixing in the example apps to get them working.

https://user-images.githubusercontent.com/1113/230191386-f0f635f2-1008-4e1f-91e2-517a65b66e31.mov

This PR now also incorporates #513 because I was having issues with pnpm locally without that.


packages/solid-keyring/src/providers/Keyring.ts
```diff
@@ -31,7 +36,7 @@ const defaultState: KeyringContextState = {
   space: undefined,
   spaces: [],
   agent: undefined,
-  account: undefined
+  account: window.localStorage.getItem(W3UI_ACCOUNT_LOCALSTORAGE_KEY) ?? undefined
```
gobengo, line 39: Can we make this a property on `<KeyringProvider>` so we can abstract away any localStorage details?
user of keyring can store this state however they want, as long as they pass it as something like `<KeyringProvider account={getItem(W3UI_ACCOUNT_LOCALSTORAGE_KEY)}>...`

travis: if we did that we'd need to require users of this library to also hand us a setter, so it would be something like:

```
<KeyringProvider account={getItem(W3UI_ACCOUNT_LOCALSTORAGE_KEY)} setAccount={(account) => {window.local.setItem(W3UI_ACCOUNT_LOCALSTORAGE_KEY, account)}}>
```

I don't think this is a very friendly to the users of this library - maintaining auth state across page refreshes feels like a core part of what we're trying to offer with these components and requiring a user to think about where they want to store the currently logged in account feels like a big ask - especially for new developers who probably don't particularly know where they want to store this state.

I'd definitely be open to making this an option, but I'd prefer to wait until we have at least one use-case (internal or external) that calls for it.

gobengo: We talked about this in person and both seem fine to do something that relies on `window.localStorage`, and if that leads to an error, we'll design something better in the context of that use case. :)

examples/react/multi-file-upload/src/components/Loader.jsx
```diff
@@ -0,0 +1,25 @@
...
+        </div>
+      </div>
+    )
+  } else {
+    return <ArrowPathIcon className='animate-spin h-4 w-4' />
+  }
+}
+
+export default function Loader ({ progressStatuses, className = '' }) {
+  console.log("STATUSes", progressStatuses)
```
gobengo, line 17:
```suggestion
```

gobengo (commented):
Would love to have a review from @olizilla to weigh in on
whether its important this work in environments without the DOM API (e.g. node.js v18 LTS where using `window` will ReferenceError)

packages/uploader-core/src/index.ts
```diff
@@ -1,13 +1,17 @@
...
 import type { ConnectionView, Principal } from '@ucanto/interface'
 
+import { Link, Version } from 'multiformats'
+
 export { uploadFile, uploadDirectory } from '@web3-storage/upload-client'
 
-export type { CARMetadata, Service }
+export type { CARMetadata, Service, ProgressFn, ProgressStatus, FetchOptions }
+
+export type ProgressStatuses = Record<string, ProgressStatus>
```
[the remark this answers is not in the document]
travis: from the perspective of this library, I don't really think so - in practice it's usually going to be the URL of the upload whose progress is being reported, but all we care about is that it uniquely identifies a particular upload so that it can be distinguished from other uploads we are getting progress reports on...

packages/solid-keyring/src/providers/Keyring.ts
travis: I realized today that anything using these libraries will likely need to disable SSR, because the primary functionality we're delivering is managing client-side state. 

This manifests in errors like these that point out that you can't use `createContext` in components that render server-side - even if we did decide not to use `window.localStorage` in this lib it will never work in Node:

Screenshot 2023-04-07 at 2 53 18 PM

travis:
> whether its important this work in environments without the DOM API

per my comment above I don't think this is ever going to be possible - `createContext` is basically the core of what these libraries do (ie, maintain client-side w3up state) and that doesn't work on the server (because it doesn't make sense to manage client-side state on the server in this way)

examples/react/file-upload/src/ContentPage.jsx
```diff
@@ -2,10 +2,11 @@ import React, { useState } from 'react'
 import { useUploader } from '@w3ui/react-uploader'
 import { withIdentity } from './components/Authenticator'
 import './spinner.css'
+import Loader from './components/Loader'
 
 export function ContentPage () {
-  const [{ storedDAGShards }, uploader] = useUploader()
-  const [file, setFile] = useState(null)
+  const [{ storedDAGShards, progressStatuses }, uploader] = useUploader()
```
Gozala, line 8: Nit: I find plural naming confusing personally, I would probably called it uploadProgress instead personally

examples/react/file-upload/src/components/Loader.jsx
```diff
@@ -0,0 +1,24 @@
+function StatusLoader ({ progressStatus }) {
+  const { total, loaded, lengthComputable } = progressStatus
+  if (lengthComputable) {
+    const percentComplete = Math.floor((loaded / total) * 100)
+    return (
```
Gozala, line 5: Nit: Why not use native [progress element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/progress) you could even style it as you’re pleased.

examples/react/multi-file-upload/src/components/Loader.jsx
Gozala, line 1: Nit: This looks identical to file-upload/components/Loader.jsx can’t we have a shared module or something ?

packages/uploader-core/src/index.ts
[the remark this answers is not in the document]
Gozala:
```suggestion
/**
  * Upload identifier that can be used to distinguish it from other uploads in progress.
  * In practice this will be an upload URL if it is unique per upload.
  */
export type UploadID = string
/**
 * Mapping of upload identifiers to corresponding upload status.
 */
export type ProgressStatuses = Record<UploadID, ProgressStatus>
```

Given the comments above I suggest above changes

Gozala (approved):
Provided some optional suggestions. Looks good to me with and without addressing them.

examples/react/file-upload/src/components/Loader.jsx
travis: oh wow I just literally did not know about this, wild! I'll take a quick pass at this - I suspect this is probably a better project for a future pass, but if it works pretty seamlessly I like the idea of modeling this

examples/react/file-upload/src/ContentPage.jsx
travis: ahh, I agree and like that name - should be pretty quick to change since it's only in this codebase, thanks!

examples/react/multi-file-upload/src/components/Loader.jsx
travis: we couuuuuld but then we'd be in the position of publishing libraries that are solely for use in our examples - I think that for example apps like this the copy-paste you have to do when updating this component is slightly less annoying than managing a shared module.

I also think that going this direction make it a bit easier for people who copy/paste the template to create a new app, so I'd prefer to stick with this for now.

examples/react/file-upload/src/components/Loader.jsx
travis: ok I looked into this - it's cool and I'd like to use it in future work, but my attempts to style it didn't go as well as I'd hoped so I'm going to leave this one for future work, great suggestion tho

PR #500, opened 2023-04-06
feat: get w3console looking acceptable on mobile

travis:
Plenty of polish we could still do, but this takes care of the most egregious problems on smaller screens:

https://user-images.githubusercontent.com/1113/230239853-c93deef4-f8fd-426c-8002-1a8ff192cd5c.mov

We're using HeadlessUI's Dialog and Transition components to make the small-screen sidebar feel nice.


gobengo (approved):
📱

PR #502, opened 2023-04-06
feat: add warnings about the public and permanent nature of w3up uploads

### LINKED ISSUE #464: Warnings about public and permanent data
we should leave these warnings in our deployed w3consoles. These are on the NFT.Storage and web3.storage web apps and docs, and still apply here.

> ⚠️❗ __Public Data__ 🌎: All data uploaded to w3up is available to anyone who requests it using the correct CID. Do not store any private or sensitive information in an unencrypted form using w3up.

> ⚠️❗ __Permanent Data__ ♾️: Removing files from w3up will remove them from the file listing for your account, but that doesn’t prevent nodes on the decentralized storage network from retaining copies of the data indefinitely. Do not use w3up for data that may need to be permanently deleted in the future.

travis:
we'll probably want to rework this in the future, but I think this probably looks good enough for now:

Screenshot 2023-04-07 at 1 47 15 PM
Screenshot 2023-04-07 at 1 47 29 PM
resolves #464

### DIFF
--- a/examples/react/w3console/src/components/Uploader.tsx
+++ b/examples/react/w3console/src/components/Uploader.tsx
@@ -81,23 +81,26 @@ const UploaderForm = (): JSX.Element => {
           <UploaderContents />
           {hasFile ? '' : <span>Drag files or Click to Browse</span>}
         </div>
-        <div className='flex flex-col space-y-4 mt-4 text-center lg:w-1/2 lg:text-left lg:mt-0'>
-          <div className=''>
-            <h4 className='text-sm mb-2'>🌎&nbsp;&nbsp;Public Data</h4>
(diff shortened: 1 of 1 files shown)


dchoi27:
thanks! wdyt about making the warnings right justified instead of left?

travis:
works for me! will update

travis:
> could also consider putting the warnings outside the dotted line?

I like this even more! I think in this case we should just keep it below the upload box in both cases:

Screenshot 2023-04-07 at 1 47 15 PM
Screenshot 2023-04-07 at 1 47 29 PM

PR updated with this change!

gobengo (approved):
Looks great, and I'm glad we are making this super clear.

PR #503, opened 2023-04-07
fix: update @web3-storage/access dep to fix email + issue

travis:
trying to auth with an email with a + in it currently fails, upgrade to the latest version of @web3-storage/access which fixes that.

### DIFF
--- a/packages/keyring-core/package.json
+++ b/packages/keyring-core/package.json
@@ -33,7 +33,7 @@
   "dependencies": {
     "@ucanto/interface": "^6.2.0",
     "@ucanto/principal": "^5.1.0",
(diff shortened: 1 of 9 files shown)


gobengo:
> Trying to auth with an email with a + in it currently fails,

Fails how?
This way or another way? https://github.com/web3-storage/w3cli/issues/73

travis:
yep, same way as web3-storage/w3cli#73
<https://github.com/web3-storage/w3cli/issues/73>

gobengo (approved):
> While I was in here I did a bit of refactoring so that only keyring-core depends on `@web3-storage/access`

love this!

PR #507, opened 2023-04-11
feat: w3console cleanup and updates to packages to support it

travis:
Integrate some learnings from creating a new app from the ground up with some tasks I've been meaning to do for a little while to clean up w3console. This is mostly either:
a) extracting components to their own files OR
b) lifting up types to eliminate 2-level-deep dependencies (ie, allowing w3console to only depend on the `@w3ui/react-*` packages rather than pulling in `@w3ui/*-core` and `multiformats`)

**NOTE FOR REVIEWERS:**

I'm particularly curious whether this change:

https://github.com/web3-storage/w3ui/pull/507/files#diff-b93f102a72fd250d9c7b9411fa0a0e4cb24b5dbe12396017aea5405cecdc371bR9

feels right to reviewers - a user of this library was recently confused by the `Link` type and asked where they could get the `CID` - I think most users of this library will be happier to get a `CID` type here but I'm ignorant of what implications this might have for return values from the service...


examples/react/w3console/src/components/SpaceRegistrar.tsx
```diff
@@ -0,0 +1,41 @@
+import { useState } from 'react';
+import { useKeyring } from '@w3ui/react-keyring';
+import Loader from './Loader';
```
alanshaw, line 3: We should have a linter enforce the use of semicolons or not, right now things are a mix.

alanshaw (approved):
LGTM

examples/react/w3console/src/components/SpaceRegistrar.tsx
travis: great call - filed an issue here: https://github.com/web3-storage/w3ui/issues/510

PR #509, opened 2023-04-11
feat: remove space creator from sidebar

travis:
this gives us a soft restriction on the number of spaces a user can create - we'll enforce this server-side in https://github.com/web3-storage/w3up/issues/704


Gozala (approved):
Looks good to me

alanshaw (approved):

PR #513, opened 2023-04-29
fix: remove playground example

travis:
since we deprecated and removed the `@w3ui/react` project it doesn't make sense to keep playground around - it was mainly there to help develop and test the "customizable" components that we're no longer using


olizilla:
What about we reframe the playground as an alternative to codesandbox. Ideally we'd get codesandbox working again, but I think that may involve switching back from pnpm to npm... needs investigating...

If we can't get codesandbox working, then we should publish similar examples as storybooks.

Gozala (approved):
Looks good to me. We can always revert this if we decide so

travis:
> What about we reframe the playground as an alternative to codesandbox. Ideally we'd get codesandbox working again, but I think that may involve switching back from pnpm to npm... needs investigating...
> 
> If we can't get codesandbox working, then we should publish similar examples as storybooks.

so I thiiiiink codesandbox does work with pnpm now (https://codesandbox.io/examples/package/pnpm) but generally agree - I'd like to remove this for now to fix the build issues I'm seeing, but am looking into the codesandbox examples now and will file an issue with this context if not - I do think Storybook would still be a useful way to show off even the headless component examples, but that probably deserves more time than we'd like to allocate to w3ui at the moment...

PR #514, opened 2023-05-03
feat: update to latest version of ucanto and @web3-storage libraries

travis:
delightfully simple


Gozala (approved):
Wow love how straight forward it end up being.

Issue #295, opened 2023-01-26
keyring-core and the providers duplicate much of w3up-client


olizilla:
What if we significantly simplify our UI layer as minimal wrappers around w3up-client? We may sacrifice some modularity, but would reduce the surface area of what we support (and repeat) here in w3ui.

For context, the original decisions for what code lives in the headless components were made at a time when the w3up-client didn't have any of the functionality we needed. It now manages spaces and and "current space" state internally, which we duplicate here in keyring-core.

As an example... for #293 to export ucan delegations for a space I tried to save time and just copy the implementation from [`w3cli`](https://github.com/web3-storage/w3cli/blob/65f7f880f35d429acc80ff3c8cc86ee8ddb30754/index.js#L223-L239) which use the [`createDelegation` method on `w3up-client`](https://github.com/web3-storage/w3up-client/blob/8ac4a5edf20ba68b345c6073a51bde74833410c9/src/client.js#L183-L191). But our ui base doesn't provide an instances of `w3up-client`. So i could go grab the lower level [`delegations` method from the access-client agent](https://github.com/web3-storage/w3protocol/blob/a1d5ecfb8a0cedc2bbd609907efab27297256fa8/packages/access-client/src/agent.js#L521-L549), but we don't expose the agent from keyring core. I will expose a method in keyring-core to do delegations, and chase it through react-keyring, so i can use it in w3console, but it also exposes that we might have more layers to w3ui than we can afford to maintain right now, and a review of our layering would be prudent.

_aside: The name `keyring-core` should also be reviewed... it's a wrapper for an agent and spaces rather than a general key management thing._

travis:
+1 reducing layers and eliminating duplicative functionality!

The one note I'll add is that it would be nice to be able to opt out of (or at least plug into) the state management stuff in these libraries - UI-level code often wants its own state management semantics and I stubbed my toe a couple times on the spaces/currentSpace state being stored and managed inside our library when I really wanted something at the React level to handle it. 

Concrete example: when I was working on the space creator, originally I wanted to have the entire space creation process happen in the left nav, but as soon as I called `createSpace` the right-hand-side Space UI changed because `createSpace` changes `currentSpace`. In this particular case I think the workaround is better than my original plan, but this was surprising behavior at the time and I think there are many other UI decisions that might be more complicated if we aren't careful about how much state management we do under the hood of our library without giving developers better hooks into state changes, etc.

PR #517, opened 2023-05-17
feat: implement `uploadCAR` in uploader

alanshaw:
This exposes the `upload-client`'s `uploadCAR` function in the `UploaderProvider`.

I need to be able to access this to upload merkle clock events generated by bucket mutations to w3up.

### DIFF
--- a/packages/react-uploader/src/providers/Uploader.tsx
+++ b/packages/react-uploader/src/providers/Uploader.tsx
@@ -9,7 +9,8 @@ import type {
 import React, { useContext, createContext, useState } from 'react'
 import {
   uploadFile,
-  uploadDirectory
+  uploadDirectory,
+  uploadCAR
 } from '@w3ui/uploader-core'
 import { useKeyring } from '@w3ui/react-keyring'
 import { add as storeAdd } from '@web3-storage/capabilities/store'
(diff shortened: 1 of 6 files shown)


packages/react-uploader/src/providers/Uploader.tsx
```diff
@@ -117,6 +121,36 @@ export function UploaderProvider ({
...
+      if (agent == null) throw new Error('missing agent')
+
+      const storedShards: CARMetadata[] = []
+      setStoredDAGShards(storedShards)
+
+      const conf = {
+        issuer: agent,
+        with: space.did(),
+        audience: servicePrincipal,
+        proofs: await getProofs([
```
travis, line 136: If I recall correctly you don't actually need to include these proofs manually - `access-client` should look through its list of proofs and include relevant proofs automatically. Could be misremembering so worth testing....

travis (approved):
I think it's worth checking if manual proof inclusion is necessary, but can always do that post-merge - this looks great!

packages/react-uploader/src/providers/Uploader.tsx
alanshaw: Hmm, I thought that was `w3up-client`? Remember we're using `upload-client` here...

travis: oh right - what I'm thinking of is actually in `agent.js` in `access-client` which is I think what `getProofs` ends up calling into - what you've got is exactly right!

this is probably another argument in favor of (refactoring this to use `w3up-client` rather than using the upload and access clients separately...)[https://github.com/web3-storage/w3ui/issues/295]

PR #528, opened 2023-07-18
fix: use unified endpoint for staging services

travis:
now that we've migrated to using a single service, update the w3console configs to point at it


travis:
this is a very minor change and blocking testing the staging service, so I'm going to merge now but would love post-merge review - happy to make updates based on changes!

PR #530, opened 2023-07-21
feat!: upgrade access client, ucanto and other packages

travis:
We needed to upgrade the access client so that it will stop pointing at `access.web3.storage` and that required ucanto upgrades.

Bumping the major version since I'm not sure this is backwards compatible.

I've disabled the playwright tests for the vue examples for now because they were failing in test but not when the example is run manually - upgrading Vue might help but I'd like to do that in followup work.

I've also disabled codesandbox for now because it was failing, possibly because of an unsupported pnpm version - I'm hoping this will simply go away whenever codesandbox upgrades pnpm.


travis:
Have filed https://github.com/web3-storage/w3ui/issues/531 for re-enabling Vue playwright tests and fixing the codesandbox builds.

alanshaw (approved):

PR #544, opened 2023-08-28
chore: move to a single-pull-request release process

travis:
The release process for this library is currently extremely cumbersome, requiring as many as 13 different release PRs merged in dependency order (with no CI-level enforcement and broken packages very possible) and a very slow build process. Even with relatively minor changes that affect half the packages I've spent an entire day half-distracted by running this process in the background.

This PR makes it so that all updated packages will be released at the same time. This is effectively what we do now, and there's very little downside to doing this (except perhaps a bit of extra churn on version numbers on packages that didn't strictly need to be released, but arguably we should be releasing any package number bumps immediately for good repo hygiene).

Open to not doing this, but opening to at the very least start the conversation.


alanshaw (dismissed):
Any different thoughts post https://github.com/web3-storage/w3ui/pull/591 ? Personally I'd say yes lets do this.

Does it do the right thing with releasing core first etc.?

travis:
ah hm good question - I think it would set all of the version numbers and dependencies to the latest versions, so it shouldn't matter which order they're released in? but I could be missing some order-dependent thing...

Peeja:
…Yes, please? 😄

Peeja (approved):
Let's just do this.

PR #550, opened 2023-09-06
fix: bump keyring package versions to bring in latest keyring-core

travis:
this fixes an issue where we were using a broken version of access-client


travis:
self-merging b/c package bumps

PR #581, opened 2023-11-13
feat!: use w3up-client in keyring

travis:
per https://github.com/web3-storage/w3ui/issues/295 rework the keyring to use w3up-client rather than the lower level access library

I'd like to refactor these APIs soon, but for now I'm keeping them unchanged to minimize downstream changes

I am releasing this as a breaking change - I don't think anything will break, but would prefer to let downstream users opt in to this


packages/keyring-core/test/agent.spec.ts
```diff
@@ -1,23 +1,18 @@
...
+  const client = await createClient()
+  const space = await client.createSpace('test')
   expect(space).toBeTruthy()
-  expect(space.did.startsWith('did:key:')).toBe(true)
-})
-
-test('registerSpace fails if no current space is set', async () => {
-  const agent = await createAgent()
-  await expect(agent.registerSpace('foo@bar.net')).rejects.toThrowError()
+  expect(space.did().startsWith('did:key:')).toBe(true)
```
alanshaw, line 17: Rename this file?

packages/react-keyring/src/Authenticator.tsx
```diff
@@ -85,6 +85,11 @@ export const AgentLoader = ({
   return children
 }
 
+/**
+ * @deprecated use ClientLoader
```
alanshaw, line 89: This is a breaking change anyways, just remove it!

packages/react-keyring/src/providers/Keyring.tsx
```diff
@@ -81,37 +78,36 @@ export function KeyringProvider ({
...
 
   const authorize = async (email: `${string}@${string}`): Promise<void> => {
-    const agent = await getAgent()
+    const c = await getClient()
     const controller = new AbortController()
     setRegisterAbortController(controller)
 
     try {
-      await accessAuthorize(agent, email, { signal: controller.signal })
+      await c.authorize(email, { signal: controller.signal })
```
alanshaw, line 105: Did this not change to `login`?

packages/react-keyring/src/providers/Keyring.tsx
```diff
@@ -129,62 +125,65 @@ export function KeyringProvider ({
...
   }
 
-  const registerSpace = async (email: string, opts: RegisterSpaceOpts = {}): Promise<void> => {
-    const agent = await getAgent()
-    await agent.registerSpace(email, {
-      provider: opts.provider ?? (agent.connection.id.did() as DID<'web'>)
-    })
-    setSpace(getCurrentSpaceInAgent(agent))
-    setSpaces(getSpaces(agent))
+  const registerSpace = async (email: string): Promise<void> => {
```
alanshaw, line 136: Maybe this should be renamed to `provisionSpace`?

packages/react-keyring/src/providers/Keyring.tsx
```diff
@@ -129,62 +125,65 @@ export function KeyringProvider ({
...
+    const c = await getClient()
+    const account = useAccount(c, { email })
+    const space = c.currentSpace()
+    if (account && space) {
+      await account.provision(space.did() as DIDKey)
+      setSpace(c.currentSpace())
+      setSpaces(c.spaces())
+    }
```
alanshaw, line 144: Need to throw if missing account or space.

packages/react-keyring/src/providers/Keyring.tsx
```diff
@@ -129,62 +125,65 @@ export function KeyringProvider ({
...
   const setCurrentSpace = async (did: DID): Promise<void> => {
-    const agent = await getAgent()
-    await agent.setCurrentSpace(did as DID<'key'>)
-    setSpace(getCurrentSpaceInAgent(agent))
+    const c = await getClient()
+    await c.setCurrentSpace(did as DID<'key'>)
+    setSpace(c.currentSpace())
   }
 
   const loadAgent = async (): Promise<void> => {
```
alanshaw, line 153: Rename `loadClient`?

packages/react-keyring/src/providers/Keyring.tsx
```diff
@@ -129,62 +125,65 @@ export function KeyringProvider ({
...
-    if (agent != null) return
-    await getAgent()
+    if (client != null) return
+    await getClient()
   }
 
   const unloadAgent = async (): Promise<void> => {
```
alanshaw, line 158: Rename `unloadClient`

packages/react-keyring/src/providers/Keyring.tsx
```diff
@@ -129,62 +125,65 @@ export function KeyringProvider ({
...
     setSpace(undefined)
     setSpaces([])
-    setIssuer(undefined)
-    setAgent(undefined)
+    setClient(undefined)
     setAccount(undefined)
   }
 
   const resetAgent = async (): Promise<void> => {
```
alanshaw, line 165: Rename `resetClient`?

alanshaw (changes requested):
Looks good so far!

travis:
@alanshaw re: a lot of this - I'd like to avoid changing the API here so we don't have to go update all the examples. I hate the idea of trying to design the next version of this API under pressure of launching, so I propose we make these changes without any changes to the interfaces these modules export and then fast follow with a breaking change that mostly just renames various functions (but also potentially reworks/deletes/adds functions that makes sense)

what do you think?

packages/react-keyring/src/providers/Keyring.tsx
travis: HM you're right - it looks like `authorize` wasn't removed, but I do think this should use login - will update!

alanshaw (approved):

PR #582, opened 2023-11-14
fix: correctly clear local storage

travis:
when unloadAgent is called

### DIFF
--- a/packages/react-keyring/src/providers/Keyring.tsx
+++ b/packages/react-keyring/src/providers/Keyring.tsx
@@ -82,7 +82,7 @@ export function KeyringProvider ({
   connection
 }: KeyringProviderProps): JSX.Element {
   const [agent, setAgent] = useState<Agent>()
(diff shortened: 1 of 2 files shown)


packages/react-keyring/src/providers/Keyring.tsx
```diff
@@ -82,7 +82,7 @@ export function KeyringProvider ({
   connection
 }: KeyringProviderProps): JSX.Element {
   const [agent, setAgent] = useState<Agent>()
-  const [account, setAccount] = useLocalStorageState<string>(W3UI_ACCOUNT_LOCALSTORAGE_KEY)
+  const [account, setAccount, { removeItem: unsetAccount }] = useLocalStorageState<string>(W3UI_ACCOUNT_LOCALSTORAGE_KEY)
```
olizilla, line 85: https://github.com/astoilkov/use-local-storage-state#uselocalstoragestatekey-string-options-localstorageoptions

olizilla (approved):

PR #588, opened 2023-11-21
chore: remove console

alanshaw:
Console moved to https://github.com/web3-storage/console

w3ui is currently publishing an old version of console to staging @ https://staging.console.web3.storage/


travis:
There are two big things in this PR:

1) removing the deployment logic so we don't push this to staging.console.web3.storage
2) removing the w3console example from the repo

(1) is definitely a good idea, let's ship it asap
re: (2): 
I could be convinced here, but I like having this example here as a good full-on example of how to use the components together to create a basic app. Much of the functionality is reproduced in the new standalone console repository, but that is complicated by concerns more specific to our business (ie, billing) and I think it doesn't do as great a job of just showing off how these components can all work together.

I do think it's potentially worth winnowing down the set of examples we provide here, but if we decide to do that I'd consider going the other direction and removing the examples that use each of the components in isolation - I think it's more useful to see them working together than alone.

Definitely hold this opinion pretty lightly, so if you feel strongly about this I can get behind it!

alanshaw:
Ok yes I think I half agree. Lets defer the decision and just merge the CI changes.

PR #589, opened 2023-11-21
feat!: pin solid and vue upload packages

travis:
We made major releases of their dependencies, so we should make major releases of these so that someone installing one of them in isolation gets the latest keyring and core deps.


travis:
I'm just gonna merge this and then move the deps back to ^ but would love to know if there are better ways to do this, please feel free to leave feedback!

PR #590, opened 2023-11-21
chore: unpin deps

travis:
undo https://github.com/web3-storage/w3ui/pull/589


alanshaw:
Lets leave these - I don't anticipate further development in the old modules.

PR #595, opened 2023-11-29
feat: add a logout function 

travis:
This resets the store, deleting the local agent's private keys and saved delegations. This is better than the previous implementation, and is appropriate for use cases like using console on a public or shared computer.

This PR also has a small tweak to replace JSX.Element with ReactNode, which I've found works better in downstream applications.

### DIFF
--- a/packages/react/src/providers/Provider.tsx
+++ b/packages/react/src/providers/Provider.tsx
@@ -4,8 +4,7 @@ import type {
   ContextActions,
   ServiceConfig,
   Space,
-  Account,
-  Store
+  Account
 } from '@w3ui/core'
 
 import React, { createContext, useState, useContext, useEffect, ReactNode } from 'react'
@@ -51,7 +50,6 @@ export function Provider ({
   const [events, setEvents] = useState<EventTarget>()
   const [accounts, setAccounts] = useState<Account[]>([])
   const [spaces, setSpaces] = useState<Space[]>([])
-  const [store, setStore] = useState<Store>()
 
   useEffect(() => {
     if ((client === undefined) || (events === undefined)) return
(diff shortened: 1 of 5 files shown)


packages/react/src/providers/Provider.tsx
```diff
@@ -61,22 +65,32 @@ export function Provider ({
...
+    setClient(undefined)
+    setEvents(undefined)
+    setAccounts([])
+    setSpaces([])
+    // set state up again
+    await setupClient()
   }
 
-  useEffect(() => { void getClient() }, []) // load client - once.
+  useEffect(() => { void setupClient() }, []) // load client - once.
```
gobengo, line 90: if props.connection changes, wouldn't we want this to run again? so that a new client gets created and passed to `setClient`?

maybe `setupClient` assignment should use `React.createMemo` with appropriate dependencies, and this `useEffect` can depend on the variables it depends on e.g. `setupClient`

gobengo: also maybe we should have the lint rule for react hook deps?  https://github.com/facebook/react/blob/main/packages/eslint-plugin-react-hooks/README.md

travis: ah, I think you're right! I think that's orthogonal to the purpose of this PR though - happy to file an issue and even work on it today, but want to get this PR merged and shipped to fix the regression in production

gobengo, line 90:
```suggestion
  useEffect(() => { void setupClient() }, [servicePrincipal, connection, setupClient]) // load client - once.
```

gobengo (approved):
consider this suggestion https://github.com/web3-storage/w3ui/pull/595/files#r1411076680
but IMO fine to address my comment in another issue

packages/react/src/providers/Provider.tsx
travis: hmmmm, not sure about this - on at the very least needs an update to the comment to explain why we're doing something so unusual, and I'm pretty sure this would trigger the linter you suggested above (though tbh I'm not a huge fan of that linter since it is not always correct and triggers on some valid use-cases)

PR #602, opened 2023-12-01
chore: extract and test the datamodel logic from `Provider`

travis:
One of the core goals here was to fix a bug where the client wouldn't get reset when servicePrincipal or connection changed. The easiest way to make sure this was tested was to extract all of the datamodel creation and management logic to a dedicated hook and test that.

Thanks to @gobengo for pointing out the bug and suggesting the fix: https://github.com/web3-storage/w3ui/pull/595#discussion_r1411060383


packages/react/src/hooks.ts
```diff
@@ -0,0 +1,77 @@
...
+      setSpaces(client.spaces())
+    }
+
+    events.addEventListener(STORE_SAVE_EVENT, handleStoreSave)
+    return () => {
+      events?.removeEventListener(STORE_SAVE_EVENT, handleStoreSave)
+    }
+  }, [client, events])
+
+  const logout = async (): Promise<void> => {
```
gobengo, line 62: personally I would recommend a `React.useCallback` here but i don't think it is needed for any functional reason. It might make the caller's component re-render slightly less but that can always be later. (we discussed this sync and you had reaons not to which are fine)

gobengo (approved):
amazing!!

PR #603, opened 2023-12-05
feat!: fix example, buncheefication, cleanup

travis:
I was hoping to keep these changes separate, but ran into some issues with using the new packages in examples that necessitated some build changes.

1) I've updated the `file-upload` example to use the new packages. I ran into a weird build issue trying to do this - something about `Authenticator` not being exported from the UMD builds? - and did some research into how other projects I like are built, which lead me to believe that `bunchee`, the tool SWR uses for builds (https://github.com/vercel/swr/blob/main/package.json#L94), is probably our best bet...

2) I love `bunchee` so far - it's much simpler and seems to do "the right things" when building TS into libraries suitable for use in a variety of modern JS environments. You can read more about bunchee here:

https://github.com/huozhi/bunchee

I've used `bunchee` to pull the build step into each of the packages separately, which should make it easier and more efficient to work on a single package in isolation. Because `bunchee` doesn't need a bunch of manual config, this was possible with a minimum of boilerplate and setup.

3) I've removed the old packages and a bunch of configs and docs that point to them. This made (2) easier, since I could just make a wholesale switch rather than trying to keep some stuff working with the old rollup builds and other stuff working with `bunchee`.


gobengo:
We'll probably want some kind of followup task to this to unpublish the npm packages we no longer plan to maintain? Otherwise probably people will stumble upon them via search results, try to use them, and be frustrated that they are no longer supported.

gobengo (approved):
I think this is most entirely removals and good changes to the build pipeline, and afaict has few risks to the packages that remain.

"Perfection is achieved, not when there is nothing more to add, but when there is nothing left to take away."

PR #604, opened 2023-12-08
fix: upgrade access client to incorporate fixed indexeddb store

travis:


travis:
just a version bump so I'm going to merge

PR #606, opened 2023-12-08
fix: package.json explicitly wants w3up-client@11.2.1

gobengo:
Motivation:
* explicitly opt in to https://github.com/web3-storage/w3up/releases/tag/upload-client-v12.3.2 (i think it will https://github.com/web3-storage/w3ui/pull/606/commits/129f74d1a7db388dad8660b29e34b8fc4fc5ab50#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbR3615)

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,4 +1,4 @@
-lockfileVersion: '6.0'
+lockfileVersion: '6.1'
 
 settings:
   autoInstallPeers: true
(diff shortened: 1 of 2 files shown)


pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '6.0'
+lockfileVersion: '6.1'
```
travis, line 1: are you using the latest pnpm? I think there was a weird bug in a version of pnpm that did this, and was reverted. they may have un-reverted it though so if you're on the latest pnpm I'm sure you're good! I'll check too...

travis (approved):

pnpm-lock.yaml
gobengo: @travis I'm on pnpm 8.6.1 (i just got latest) and I did a re- `pnpm install` and did not produce different lockfile

gobengo: (i upgraded to 8.12.0 and it produces identical lockfile)

PR #615, opened 2023-12-14
feat: support multi file upload and directory wrapping

travis:
Make backwards compatible API changes to support multi-file upload and directory wrapping.

Also support directory selection, wrapping the wonky builtin API with something a little cleaner.

Revive the multi file upload and uploads lists examples!

### DIFF
--- a/packages/react/src/Uploader.tsx
+++ b/packages/react/src/Uploader.tsx
@@ -36,6 +36,10 @@ export interface UploaderContextState {
    */
   file?: File
   /**
+   * Files to be uploaded
+   */
+  files?: File[]
+  /**
    * A callback that can be passed to an `onSubmit` handler to
    * upload `file` to web3.storage via the w3up API
    */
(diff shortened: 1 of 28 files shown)


packages/react/src/Uploader.tsx
```diff
@@ -150,14 +173,18 @@ export const UploaderRoot: Component<UploaderRootProps> = createComponent(
...
             dataCID,
             status,
             error,
             handleUploadSubmit,
             storedDAGShards,
             uploadProgress
           },
-          { setFile: setFileAndReset }
+          {
+            setFile: (file?: File) => { setFilesAndReset((file != null) && [file]) },
```
alanshaw, line 185:
```suggestion
            setFile: (file?: File) => { file && setFilesAndReset([file]) },
```

travis: ah yea this is what I had originally, actually, but the linter wasn't happy. now tsc isn't happy. will find something that pleases them both, lol

alanshaw (approved):

PR #617, opened 2024-01-03
fix: upgrade upload client

travis:
bring in fix for `fetch.node` issue


travis:
self merged since it's just a version upgrade

PR #633, opened 2024-06-06
feat: upgrade to latest w3up-client

alanshaw:


packages/core/package.json
```diff
@@ -34,7 +34,7 @@
     "@ucanto/transport": "^9.1.1",
     "@web3-storage/access": "^18.4.0",
```
vasco-santos, line 35: could we also update access to latest? Not sure if there are any implications there

vasco-santos (approved):

Peeja:
Closing, superseded by https://github.com/storacha-network/w3ui/pull/634

PR #634, opened 2024-08-30
feat: Use client v16.0.0 and the Blob protocol

Peeja:
Upgrades `@web3-storage/upload-client` and `@web3-storage/access` to the latest versions. Causes downstream application to use the Blob protocol transparently, without affecting the API it presents.

Supersedes #633, an older version of the same upgrade.


hannahhoward (approved):
LGTM

PR #637, opened 2024-08-30
refactor: `onSubmit` handler should have correct type

Peeja:
Because we're using `createElement` from an old version of AriaKit, TS isn't checking the type of our `onSubmit` handler. It's not currently broken, but it's not being checked either. This change ensures it's correct, and makes it more correct by satisfying the types and the linter. Notably, an async function should not be a handler directly; instead, a handler should kick off an async function and attach an error handler as required. This satisfies lint rules against leaving promises dangling (apparently) accidentally.


hannahhoward (approved):
LGTM

PR #644, opened 2024-10-03
feat: Re-`claim` delegations during client setup

Peeja:
The rest of https://github.com/storacha/project-tracking/issues/125. This along with [generating the space delegations on the fly during `access/claim`](https://github.com/storacha/w3up/pull/1555) will make new spaces appear on the page without logging out.

---

If the client is restored from local storage, it may be out of date. This fetches the latest delegations from the server and updates the client. This makes loading the client from local storage a stale-while-revalidate operation.


fforbeck (approved):
Looks good to me!

PR #645, opened 2024-10-03
chore: Add `dev` script

Peeja:
Use `pnpm dev` to build both packages in watch mode.

Also, upgrades `bunchee` to get past [a bug in its watch mode](https://github.com/huozhi/bunchee/issues/383). This is a big upgrade (two major versions), but everything builds fine—the upgrade doesn't appear to be as breaking as that sounds. Worth double-checking, though.

### DIFF
--- a/examples/react/components/package.json
+++ b/examples/react/components/package.json
@@ -26,7 +26,7 @@
     "@w3ui/react": "workspace:^"
   },
   "devDependencies": {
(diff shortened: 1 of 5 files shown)


travis (approved):
praise: this looks great! checked out the branch, ran `pnpm dev`, changed a file, rebuilt.

PR #650, opened 2024-12-04
feat: Use latest client (`@storacha/client`)

Peeja:
This should fix uploads in non-prod environments (a fix which was applied in the new client, but not the old one).


hannahhoward (approved):
LGTM

